PR #59, opened 2026-03-23
Implement RT-DETR model with training, validation, and prediction

testdummyvt:
- This pull request adds support for the RT-DETR object detection model family to the `libreyolo` codebase. 
- It introduces the new `LibreYOLORTDETR` class, implements model-specific preprocessing and output parsing, and integrates the model into the backend and model detection mechanisms. 
- Additionally, it brings in the RT-DETR model architecture and its utilities, including backbone, box operations, and training configuration.


testdummyvt:
@brohoya @EHxuban11 Tested model for ONNX and tensorRT exports as well. 

Currently, the only thing left is to incorporate RTDETR into the LibreYOLO class. 

This needs weighst to be in your huggingface repo. Currently weights are available at [https://huggingface.co/testdummyvt/libre_rtdetr_v1_weights/tree/main](https://huggingface.co/testdummyvt/libre_rtdetr_v1_weights/tree/main)

Once you guys upload it to your repo, I can complete integrating RTDETR into LibreYOLO class.

brohoya:
> @brohoya @EHxuban11 Tested model for ONNX and tensorRT exports as well.
> 
> Currently, the only thing left is to incorporate RTDETR into the LibreYOLO class.
> 
> This needs weighst to be in your huggingface repo. Currently weights are available at https://huggingface.co/testdummyvt/libre_rtdetr_v1_weights/tree/main
> 
> Once you guys upload it to your repo, I can complete integrating RTDETR into LibreYOLO class.

Hey @testdummyvt, just uploaded your weights if you wanna test: 
https://huggingface.co/collections/LibreYOLO/libreyolo-models

libreyolo/models/__init__.py
```diff
@@ -167,9 +168,11 @@ def LibreYOLO(
 
     weights_dict = _unwrap_state_dict(state_dict)
 
-    # Ensure RF-DETR is registered if its keys are present
+    # Ensure RF-DETR is registered if its keys are present, but avoid
+    # treating RT-DETR checkpoints as RF-DETR.
+    is_rtdetr = LibreYOLORTDETR.can_load(weights_dict)
```
testdummyvt, line 173: As both RF-DETR and RT-DETR share transformer arch naming, I needed to add a if condition to check if it is RT-DETR or not.

testdummyvt:
@brohoya Everything works now. It is ready for review and merge.

libreyolo/models/rtdetr/trainer.py
```diff
@@ -0,0 +1,383 @@
...
+            total_loss += loss_val
+
+            del outputs, loss
+
+            # LR update with per-group scaling
+            base_lr = self.lr_scheduler.update_lr(self.current_iter + 1)
+            for param_group in self.optimizer.param_groups:
+                param_group["lr"] = base_lr * param_group.get("lr_ratio", 1.0)
+            num_batches += 1
+
```
brohoya: The _train_epoch seems to mostly be duplicate code from the BaseTrainer method. The difference being the per group LR update with a ratio. 

Could we instead create a hook like this in BaseTrainer:
```python
def _scale_lr(self, base_lr: float, param_group: dict) -> float:
    """Hook for per-group LR scaling. Override in subclasses."""
    return base_lr
```
and override it in the RTDETRTrainer with the per group LR implementation like so:
```python
def _scale_lr(self, base_lr: float, param_group: dict) -> float:
    return base_lr * param_group.get("lr_ratio", 1.0)
```

We can then remove _train_epoch from RTDETRTrainer and update the BaseTrainer version to use the _scale_lr in the LR update:
```python
            # LR update
            lr = self.lr_scheduler.update_lr(self.current_iter + 1)
            for param_group in self.optimizer.param_groups:
                param_group["lr"] = self._scale_lr(lr, param_group)
            num_batches += 1
```

libreyolo/models/rtdetr/backbone.py
```diff
@@ -0,0 +1,267 @@
...
+        super().__init__()
+        self.conv = nn.Conv2d(
+            ch_in, ch_out, kernel_size, stride,
+            padding=(kernel_size - 1) // 2 if padding is None else padding,
+            bias=bias)
+        self.norm = nn.BatchNorm2d(ch_out)
+        self.act = nn.Identity() if act is None else get_activation(act)
+
+    def forward(self, x):
+        return self.act(self.norm(self.conv(x)))
```
brohoya, line 78: The ConvNormLayer is also defined `nn.py`, can we import it from there for the backbone ?

pyproject.toml
```diff
@@ -18,6 +18,7 @@ dependencies = [
     "opencv-python>=4.11.0.86",
     "tqdm>=4.65.0",
     "pycocotools>=2.0.0",
+    "scipy>=1.7.0",
```
brohoya: Nitpick: this is only used by RTDETR, could we add it as an optional dependency like that ?
```toml
[project.optional-dependencies]
  rtdetr = ["scipy>=1.7.0"]
```

This mirrors the rfdetr pattern. Then in loss.py, guard the import:
```python
  try:
      from scipy.optimize import linear_sum_assignment
  except ImportError:
      raise ImportError(
          "scipy is required for RT-DETR training. "
          "Install with: pip install libreyolo[rtdetr]"
      )
```

libreyolo/models/rtdetr/model.py
```diff
@@ -0,0 +1,514 @@
...
+        img_float = img_resized.astype(np.float32) / 255.0
+
+        # HWC -> CHW
+        img_chw = img_float.transpose(2, 0, 1)
+
+        # To tensor and add batch dimension
+        input_tensor = torch.from_numpy(img_chw).unsqueeze(0)
+
+        if next(self.model.parameters()).is_cuda:
+            input_tensor = input_tensor.cuda()
```
brohoya: I think the InferenceRunner already handles the device, is there another reason for this ?

brohoya (commented):
Hey @testdummyvt, looks solid overall! 
Finished reviewing the changes, I will try and test it more tomorrow.

libreyolo/models/rtdetr/trainer.py
testdummyvt: Ah, yes. Most of the code I wrote is to make sure I do not change the original code in Libreyolo. That's why I made a copy of the train epoch. 

I think what you said makes sense. I will do those changes.

pyproject.toml
testdummyvt: This is good idea, yeah. I will do that.

testdummyvt:
@brohoya, check out the new changes. 

I did update the code as per your suggestion. 

Tested both training and testing of the models.

PR #64, opened 2026-03-24
perf: speed-up annotation preload by reading image size via Pillow header

pxydi:


EHxuban11:
All 124 unit tests pass

PR #67, opened 2026-03-26
fix: preserve custom class names in checkpoints

### LINKED ISSUE #66: BUG: Custom class names not persisted in finetuned checkpoints
Custom class names from data.yaml were never propagated to the model during training.

Reported by @pxydi for LibreYOLOX in the issue https://github.com/LibreYOLO/libreyolo/issues/61

brohoya:


brohoya:
Model trainers have a lot of duplicate code, I will refactor it later. Let's fix the bug for now.

PR #181, opened 2026-05-05
Native RF-DETR port to LibreYOLO

brohoya:
## Description
Replaces the upstream rfdetr package wrapper with a native LibreYOLO integration, matching the integration shape of YOLOX, YOLO9, D-FINE, RT-DETR, EC, etc.

## Changes
- New native package libreyolo/models/rfdetr/ (15 files, ~6.7k LoC) covering: DINOv2-with-windowed-attention backbone, LW-DETR transformer + multi-scale deformable attention, Hungarian matcher + SetCriterion, segmentation head, postprocess, native trainer (RFDETRTrainer(BaseTrainer)).
- LibreRFDETR wraps it as the family-facing BaseModel, fully adapted to the new task paradigm: SUPPORTED_TASKS = ("detect", "segment"), TASK_INPUT_SIZES, task init kwarg, automatic task resolution (explicit > legacy segmentation flag > filename -seg suffix > checkpoint task metadata).
- RFDETRTrainer.on_forward adopts the new polygons parameter.
- Optional dep [rfdetr] now resolves to transformers + scipy (no more rfdetr PyPI package).
- export/onnx.py keeps RF-DETR's (pred_boxes, pred_logits) tuple order, distinct from D-FINE / DEIM / EC's (pred_logits, pred_boxes) ordering with a clarifying comment.
- tests/e2e/test_onnx.py::test_onnx_seg_export_produces_masks fixed to assert RF-DETR's pred_masks naming (pre-existing bug, unrelated to the port).

## Compatibility

- Weight-compatible with upstream Roboflow HuggingFace checkpoints, they load unmodified through LibreRFDETR("LibreRFDETR{n,s,m,l}.pt") and segmentation variants (-seg suffix).
- COCO 91-class -> YOLO 80-class output remapping preserved.
- All public surfaces (predict(), val(), export(), track()) return LibreYOLO-standard Results / Boxes / Masks.


pyproject.toml
```diff
@@ -36,7 +36,8 @@ onnx = [
     "onnxruntime>=1.16.0",
 ]
 rfdetr = [
-    "rfdetr>=1.6.2,<2.0.0",
```
Borda, line 39: Hello, could not find any description, so curious if you could share why you prefer copying code instead of using the official package? 🦝

brohoya: Hi @Borda, sorry for the missing description.
  
The goal of LibreYOLO is to put several detection families behind one LibreYOLO("…") factory with a uniform inference/validation/export/tracking API. We have been using the official rfdetr package, but it's the only (and best 🔥) family in our catalog that isn't natively integrated, and it creates friction for us every time we are adding a new feature since we need to think about the implementation for native models and the rfdetr wrapper (e.g. exports, validation, installation, etc), it is twice the work. Also, we might add pose estimation to RF-DETR later and it would be much easier with a "native" implementation.

That said, we really want to make sure that Roboflow’s work is clearly credited. We are preserving the attribution in each file and in THIRD_PARTY_NOTICES.txt, and if there’s a better way you’d like us to handle naming/notices/docs here, we are open to adjust.

🦝

Issue #152, opened 2026-04-30
Investigate why coco (json) and yolo (.txt) labels produce different evaluation results when running coco val 2017


EHxuban11:

mateo19182:
Hi! Wanted to start contributing and thought this might be a good issue to start. 

Both paths (.txt and .json) feed the same pycocotools.cocoeval.COCOeval, but build ground truth differently:

  1. iscrowd is hard-coded to 0 https://github.com/LibreYOLO/libreyolo/blob/c2189eb77cf62c1ca7992f885dc49e032e785b63/libreyolo/data/yolo_coco_api.py#L258   .txt has no crowd marker, Ultralytics has the same issue ([yolov5#7732](https://github.com/ultralytics/yolov5/pull/7732)).
  
  2. area is bbox in txt, segmentation area in json https://github.com/LibreYOLO/libreyolo/blob/c2189eb77cf62c1ca7992f885dc49e032e785b63/libreyolo/data/yolo_coco_api.py#L112 same annotation can land in different size buckets.
  
  3. DETR top-k cutoff is gated to the json path https://github.com/LibreYOLO/libreyolo/blob/c2189eb77cf62c1ca7992f885dc49e032e785b63/libreyolo/validation/detection_validator.py#L436-L438  on txt uses the user's conf_thres, which can remove the tail of the set.

1 and 2 don't really have fixes, just documenting it seems reasonable. For 3, skip the _coco_annotation_file check.

There might be other issues that I haven't found, but txt and json will never be 100% aligned as the txt contains less info.

Let me know if this makes sense and I will open a PR with the relevant changes!

mateo19182:
Ran bench on COCO val2017. Same config, only change wheter using COCO json or YOLO txt.

**`LibreYOLOXs.pt`**

| metric | json | yolo | diff |
|---|---:|---:|---:|
| `mAP50-95` | 0.3897 | 0.3853 | −0.0044 |
| `mAP50` | 0.5793 | 0.5722 | −0.0071 |
| `mAP_small` | 0.2224 | 0.1644 | **−0.0580** |
| `mAP_medium` | 0.4326 | 0.3672 | **−0.0654** |
| `mAP_large` | 0.5223 | 0.5286 | +0.0063 |
| `AR100` | 0.5602 | 0.5602 | 0 |

**`LibreDFINEs.pt`**

| metric | json | yolo | diff |
|---|---:|---:|---:|
[...]

PR #186, opened 2026-05-06
fix: align COCO eval between json and yolo label paths (#152)

mateo19182:
https://github.com/LibreYOLO/libreyolo/issues/152

- Removes _coco_annotation_file guard so DETR top-k cutoff (conf_thres=0) applies to both JSON and YOLO paths
- Adds COCO annotations download so the validator defaults to JSON path (which carries proper iscrowd, area, etc.)

### DIFF
--- a/libreyolo/config/datasets/coco-val-only.yaml
+++ b/libreyolo/config/datasets/coco-val-only.yaml
@@ -103,6 +103,12 @@ download: |
 
   dir = Path(yaml["path"])
 
+  # Download annotations (COCO JSON) - required for proper COCO evaluation
+  download(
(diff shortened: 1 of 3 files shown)


brohoya:
Thanks for the changes! 

Small follow-ups before merging: `coco.yaml` (the full COCO config) needs the same snippet, both yamls resolve to the same `<datasets>/coco/` layout, so users running `model.train(data="coco")` or `model.val(data="coco")` hit the same divergence. Same addition as in `coco-val-only.yaml`.

brohoya:
LGTM

PR #192, opened 2026-05-07
Remove the .resolve from the get_img_files method

### LINKED ISSUE #187: Data loading does not work as expected if the images dir is a symlink
Hi! Thanks for this amazing work, really appreciate it!

So, I've been trying to train a YOLOX model using a data.yaml file with .txt files for the train/test/val splits. My data directory structure is as follows:

```
<DATA DIR>
│
└───data.yaml
└───test.txt
└───train.txt
└───val.txt
└───images
    │
    └───img1.png
    └───img2.png
    └───img3.png
    └───...
└───labels
    └───img1.txt
    └───img2.txt
    └───img3.txt
    └───...
```

The data.yaml is properly structured but the images directory is actually a symlink to another directory that doesn't have the work 'images' in its absolute path (nor the substitution of 'labels' for 'images' would point to the labels). When I start the training, I can see that the images are properly loaded. However, I've seen that the images filepath are considered using their original path and not with the symbolic link. This is because of the .resolve() [here](https://github.com/LibreYOLO/libreyolo/blob/613044e423562ac084404aba4678e07a974c4f5a/libreyolo/data/utils.py#L178). Therefore, when transforming them into labels using the [image2labels_path](https://github.com/LibreYOLO/libreyolo/blob/613044e423562ac084404aba4678e07a974c4f5a/libreyolo/data/utils.py#L191) method, the labels point to a location that doesn't exist. Training kicks in but the loss rapidly goes to 0, indicating that the models are trained as if the images have no annotations.

I've removed the .resolve() and training starts as expected. Any reason why to add the resolve there? Or should there be any better handling for symbolic links?

Thanks!

KostadinovShalon:
This PR addresses the bug where using .resolve() for the images path in the get_img_files method would replace the paths generated by symlinks, therefore conflicting with the logic of replacing 'images' with 'labels' in the images path, causing not to find the labels. Closes #187

### DIFF
--- a/libreyolo/data/utils.py
+++ b/libreyolo/data/utils.py
@@ -175,14 +175,14 @@ def get_img_files(path: Union[str, Path, List], prefix: str = "") -> List[Path]:
                         # Relative to txt file's parent directory
                         img_path = path.parent / img_path
                     if img_path.exists() and img_path.suffix.lower() in IMG_FORMATS:
-                        img_files.append(img_path.resolve())
+                        img_files.append(img_path)
         return sorted(img_files)
 
     elif path.suffix.lower() in IMG_FORMATS:
         # Single image file
         if not path.exists():
             raise FileNotFoundError(f"Image file not found: {path}")
-        return [path.resolve()]
+        return [path]
 
     else:
         raise ValueError(f"Unsupported path format: {path}")


EHxuban11:
Hello @KostadinovShalon , thank you for the PR!

This fixes exactly what we were talking about in the issue. 

Merged

PR #194, opened 2026-05-07
43 add tta

imagra93:


EHxuban11:
Hello @imagra93 , this looks fine from my side. Thank you for your work!

I redirected the merge to dev (instead of main), so that we can have this in the next release.

I will run the full final benchmarks when I get my computer and before the first release

PR #197, opened 2026-05-08
Fix: Bug in preprocess of yolonas

imagra93:
**The Problem**
I noticed that validation metrics were incorrectly low (near zero). The issue was a padding mismatch: the validation step was using **top-left padding**, while the inference/post-processing logic expected **center-padding**.

This created an **~81-pixel offset** on non-square images, causing the model to miss nearly all ground truth targets.

**The Fix**
Updated `YOLONASValPreprocessor` to use center-padding, matching the `preprocess_numpy` logic exactly.

**Results**
Validation metrics are now correct:

* **$mAP_{50}$:** $0.063 \rightarrow \mathbf{0.644}$
* **$mAP_{50-95}$:** $0.019 \rightarrow \mathbf{0.469}$

### DIFF
--- a/libreyolo/validation/preprocessors.py
+++ b/libreyolo/validation/preprocessors.py
@@ -266,8 +266,6 @@ class YOLONASValPreprocessor(YOLO9ValPreprocessor):
     offset that collapses baseline mAP to near zero on non-square images.
     """
 
-    RESIZE_SIZE: int = 636
-
     @property
     def wants_unresized_image(self) -> bool:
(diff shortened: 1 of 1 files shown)


EHxuban11:
Thank you for reporting and fixing this bug. 

After some digging, it was introduced here: https://github.com/LibreYOLO/libreyolo/commit/6ba53dd8dc9b1d22c174899ba944b15ad1ecbc2c

In the future, we will run e2e tests more often.

PR #199, opened 2026-05-08
Fix: Segmentation mask rendering with augment

imagra93:
Corrects a bug preventing masks from displaying when TTA is active. Also includes a minor code cleanup for better clarity.


EHxuban11:
Good spot!

I was able to reproduce this by doing save=True to an image with and without TTA. The solution fixed it. The code looks correct. 

I launched several agents checking for regression and it looks fine.

Merged to dev!

PR #203, opened 2026-05-09
Speed up NMS: collapse per-class loops to batched_nms (#202)

### LINKED ISSUE #202: Speed up NMS
The library has a very naive python loop for nms instead of using the pytorch nms which is much faster and has a specialized cuda core and so on.

EHxuban11:
## Summary

Three call sites used a `for cls in unique_classes:` Python loop with one `torchvision.ops.nms` call per class. The loop is mathematically equivalent to a single `torchvision.ops.batched_nms(boxes, scores, classes, iou)` call — same library, same CUDA kernel, same class-offset trick — but launches N kernels and a Python iteration per unique class.

This bites worst on validation, where `detection_validator.py` forces `conf_thres=0.0` for the COCO topk eval. The conf prefilter no longer shrinks the loop, so YOLOX and YOLO9 hit the full ~80-iteration worst case on every image.

## Changes

| File | What |
|---|---|
| `libreyolo/utils/general.py` | `postprocess_detections` per-class loop → single `batched_nms`. Used by YOLOX + YOLO9. |
| `libreyolo/models/base/model.py` | TTA merge per-class loop → single `batched_nms`. |
| `libreyolo/models/base/inference.py` | Tiled-inference merge per-class loop → single `batched_nms`. |

Cleanup that landed in the same patch (load-bearing for the import surface):

- Deleted `nms()` in `utils/general.py` (76-line Python while-loop with `.item()` GPU sync). Was only invoked as a torchvision-missing fallback; torchvision is a hard dep via torch itself, so the fallback is dead.
- Deleted `postprocess_batch()` in `utils/general.py` (119-line fast batched validator path that was written but never imported anywhere).

## Audit

Verified the 11 detection families before deciding scope (sub-agent reports):

- **Slow path (per-class Python loop):** YOLOX, YOLO9. Both fixed by this PR.
- **NMS-free (top-K):** YOLO9-E2E, EC, RF-DETR, D-FINE, DEIM, DEIMv2, RT-DETR. Untouched.
- **Already fast (single `batched_nms`):** YOLO-NAS, PicoDet. Untouched.

So the patch is laser-targeted: 2 families benefit on the inference path, all 11 benefit on the TTA + tiled paths.

## Diff

`+30 / -257` across 3 files.

## Test plan

- [x] `pytest tests/unit/ -q` — 680 passed, 37 skipped, 0 failed
- [ ] e2e/RF1 in CI
- [ ] Spot-check val mAP unchanged on YOLOX or YOLO9 (math is bit-equivalent so this should be a formality)

Closes #202.

### DIFF
--- a/libreyolo/models/base/inference.py
+++ b/libreyolo/models/base/inference.py
@@ -16,7 +16,7 @@ from typing import TYPE_CHECKING, Dict, Generator, List, Optional, Tuple, Union
 import torch
 
 from ...utils.drawing import draw_boxes, draw_keypoints, draw_masks, draw_tile_grid
-from ...utils.general import get_safe_stem, get_slice_bboxes, nms, resolve_save_path
+from ...utils.general import get_safe_stem, get_slice_bboxes, resolve_save_path
 from ...utils.image_loader import ImageInput, ImageLoader
 from ...utils.predict_args import normalize_predict_kwargs
 from ...utils.results import Boxes, Keypoints, Masks, Results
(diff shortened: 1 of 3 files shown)


EHxuban11:
Too dangerous to this change without having the capability of running every e2e.

PR #207, opened 2026-05-11
add gradient accumulation

imagra93:
# feat(training): gradient accumulation support

Adds a `grad_accum_steps` config field that lets any model family simulate a
larger effective batch size by accumulating gradients over multiple forward
passes before each optimizer step. Default is `1`, so existing behaviour is
fully preserved.

### `TrainConfig` (`libreyolo/training/config.py`)

New field:

```python
grad_accum_steps: int = 1
```

Pass it as any other training kwarg:

```python
trainer.train(batch=4, grad_accum_steps=2)  # effective batch = 8
```

### `BaseTrainer._train_epoch` / `DEIMTrainer._train_epoch` / `DFINETrainer._train_epoch`

All three `_train_epoch` implementations received the same set of changes:

**Loss scaling**

Each micro-batch loss is divided by `actual_window` before `.backward()`:

```
loss = outputs["total_loss"] / actual_window
```

`actual_window` is `min(accum, remaining_batches)` — it shrinks for the last
window of the epoch when `len(train_loader)` is not divisible by `accum`. This
ensures the gradient contribution of every sample is weighted identically
regardless of where it falls in the epoch; without it the final partial window
would contribute inflated gradients relative to full windows.

For logging, `loss_val` is restored to the un-scaled value (`loss.item() * accum`)
so the displayed loss is directly comparable across runs with different `accum`
values.

**Optimizer and scaler step timing**

Previously the optimizer stepped on every batch. Now it only steps when
`is_opt_step` is true:

```
is_opt_step = (batch_idx + 1) % accum == 0 or batch_idx == len(train_loader) - 1
```

`optimizer.zero_grad(set_to_none=True)` is moved to the *start* of each
accumulation window (when `batch_idx % accum == 0`) instead of just before the
backward pass. This is the standard pattern — zeroing inside the window would
discard the accumulated gradients.

**EMA update timing**

EMA was previously updated on every batch. It now updates only on optimizer
steps (`if is_opt_step`). This keeps EMA semantics consistent: one EMA step per
weight update, regardless of how many micro-batches produced that update.

**LR scheduler step timing**

Same move — `lr_scheduler.update_lr()` is now called only on optimizer steps.
With `accum=1` this is identical to before. With `accum=N` the LR advances once
per N batches, which is correct: the schedule should track weight updates, not
data passes.

**`current_iter` semantics**

`current_iter` now counts **optimizer steps**, not batch indices:

```
opt_step = epoch * steps_per_epoch + batch_idx // accum
self.current_iter = opt_step
```

Any downstream code (e.g. schedulers, logging) that reads `current_iter` will
see a consistent step count regardless of accumulation factor.

### Scheduler receives the right `steps_per_epoch`

`create_scheduler(iters_per_epoch)` is called with `ceil(N / accum)` instead of
the raw `len(train_loader)`. This means the warm-up duration, cosine period, and
any other schedule features are expressed in optimizer steps, not batch steps.

**Practical consequence:** a 5-epoch warm-up with `accum=4` will span the same
number of real weight updates as with `accum=1`. Without this fix, the warm-up
would be 4× shorter in real terms when accumulation is used, causing a much
steeper LR ramp.

### tqdm suppressed outside a TTY (DEIM and D-FINE trainers)

```python
disable=not sys.stderr.isatty()
```

Prevents tqdm from spamming log files when training is launched non-interactively
(e.g. via `nohup`, a scheduler, or CI). `BaseTrainer` already had equivalent
logic; this brings the family-specific trainers in line.

## Effective batch size and LR

The code uses a linear scaling rule (`effective_lr = lr0 * batch / 64`). When
using accumulation, `batch` should be set to the **per-GPU micro-batch size**,
and `lr0` should be set relative to the *effective* batch:

| Scenario | `batch` | `accum` | effective batch | `lr0` |
|---|---|---|---|---|
| Baseline | 8 | 1 | 8 | `x` |
| Accum ×2 | 4 | 2 | 8 | `2x` (to keep effective_lr equal) |
| Accum ×4 | 4 | 4 | 16 | `x` (effective_lr doubles — intended) |

The comparison script `scripts/compare_grad_accum.py` verifies that
`batch=8/accum=1` and `batch=4/accum=2` produce the same `effective_lr` and
converge to a similar loss trajectory across YOLOX, DEIM, and D-FINE.

## No breaking changes

With `grad_accum_steps=1` (the default) all three code paths are identical to
the previous behaviour: every batch is an optimizer step, `zero_grad` is called
once per batch, EMA and LR advance every batch, and `current_iter` increments
by one per batch. The only observable difference is that `zero_grad` now uses
`set_to_none=True`, which is a minor memory optimisation with no effect on
correctness.


imagra93:
## Verification scripts for gradient accumulation

Two helper scripts under `scripts/` to validate the accumulation implementation:

- **`download_mask_wearing.py`** — downloads the public `mask-wearing-608pr` dataset (CC BY 4.0, 105 train images, 2 classes: mask / no-mask) from the LibreYOLO HuggingFace org.
- **`compare_grad_accum.py`** — fine-tunes YOLOX-s from pretrained COCO weights under two configurations with the same effective batch size:

| Run | `batch` | `grad_accum_steps` | Optimizer steps/epoch |
|-----|---------|-------------------|----------------------|
| A   | 32      | 1                 | 4                    |
| B   | 2       | 16                | 4                    |

To reproduce:

```bash
bash scripts/compare_grad_accum.sh --device cuda
```

Output is written to `test_gradient_acc.txt`. On a 50-epoch run, mean absolute loss difference was **0.168**, with both runs reaching mAP50 ≈ 0.60–0.70.

> The small residual gap is expected: BatchNorm statistics are computed per micro-batch (size 2), not over the effective batch (size 32) — a known inherent limitation of gradient accumulation in BN-heavy architectures, not a correctness bug.

These scripts will be removed before merge.

---

## Open question: API design for accumulation

Currently accumulation is controlled via two explicit parameters:

```python
model.train(data=..., batch=2, grad_accum_steps=16)
```

An alternative is an Ultralytics-style `nbs` (nominal batch size) parameter, where accumulation steps are derived automatically:

```python
grad_accum_steps = nbs // batch
```

A user who can only fit `batch=8` in memory but wants the effective behaviour of `batch=64` would simply set `nbs=64` — no mental arithmetic required.

| Approach | Pros | Cons |
|----------|------|------|
| `grad_accum_steps` | Explicit, predictable | Requires user to calculate steps manually |
| `nbs` | User-friendly for memory-constrained cases | Less transparent about what's happening |

Worth deciding before this lands. Happy to implement either direction.

EHxuban11:
Merged your code with correct atribution but in this PR: https://github.com/LibreYOLO/libreyolo/pull/236

PR #209, opened 2026-05-11
Fix 80 classes hardcoded "bug"

atunSensei:
This should fix the "we have 80 classes if you don't put nc: in data.yaml" bug when training YOLO9.

### DIFF
--- a/libreyolo/models/base/model.py
+++ b/libreyolo/models/base/model.py
@@ -396,10 +396,19 @@ class BaseModel(ABC):
                         )
 
                 ckpt_nc = loaded.get("nc")
+                ckpt_names = loaded.get("names")
+
+                # Infer nc from names if missing from checkpoint
+                if ckpt_nc is None and ckpt_names is not None:
+                    ckpt_nc = len(ckpt_names)
+
+                # Infer nc from existing tensor detect_nb_classes
+                if ckpt_nc is None and hasattr(self, "detect_nb_classes"):
+                    ckpt_nc = self.detect_nb_classes(state_dict)
+
                 if ckpt_nc is not None and ckpt_nc != self.nb_classes:
                     self._rebuild_for_new_classes(ckpt_nc)
 
-                ckpt_names = loaded.get("names")
                 effective_nc = ckpt_nc if ckpt_nc is not None else self.nb_classes
                 if ckpt_names is not None:
                     self.names = self._sanitize_names(ckpt_names, effective_nc)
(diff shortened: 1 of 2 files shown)


EHxuban11:
Hello @fegonzalezg , What is this bug? can you show me how to reproduce it, please?

atunSensei:
When you train YOLO with a data.yaml file that does _not_ have "nc", we straight up assume 80 classes without even counting the names/classes.
The resulting model will show 80 classes. If you have, say 30 classes, the model will "fill up" the empty classes all the way to 80.
I saw this on the ONNX export using Netron. 80 seemed like a very specific number, so I left it be, but added an extra check for the names array to count them before defaulting to 80.

I saw that the docs does indeed ask for "nc:", but when I export datasets in "YOLO format", we don't get the "nc:" by default.

Magellol:
> 80 seemed like a very specific number, so I left it be

I think 80 comes from the COCO dataset, which I believe the original YOLO was trained against.

atunSensei:
You're right. Then 80 default makes sense. What doesn't make sense is that we assume 80 classes only because one forgets to put "nc: numberOfClasses". We should at least try to count the classes. Right?

atunSensei:
Hi @EHxuban11, is there any specific scenarios you'd like me to test to merge? The change only adds a check whenever there's no "nc: numberOfClasses" in the yaml, and does a quick sum of the classes array in the yaml. If there are no classes at all in the yaml, then the code will proceed as always and assume the hardcoded 80 classes.

UL's yaml doesn't require user to write the total number of classes. So if the aim to is to make it as seamless as possible for someone coming over from the UL universe, then this change is beneficial.

Currently, if one does NOT explicitly write the number of classes, then the code defaults to 80 classes.

EHxuban11:
Hello @fegonzalezg ,

Sorry, I've been quite busy. You are right regarindg that when not defining nc, 80 is used instead of counting. Also, having to define nc differs from UL as you say, so it needs fixing. You are fixing for YOLO9, can you try to fix it for every model? 

It shouldn't be too hard according to Claude Code:

` Yes, very simple. 2 lines in a single place (libreyolo/data/utils.py), right before the return config at line 309:

  if config.get("nc") is None and config.get("names") is not None:
      config["nc"] = len(config["names"])

  That's it. All 13 trainers read data_config.get("nc") from the dict returned by load_data_config, so this centralized
   inference makes them all behave correctly without touching any trainer.`

Regarding the _load_weights changes, they are a different scope from the data.yaml bug, please split it into another PR.

Great bug report and also I like that you understand the goal of maintaining the same API as UL, we are very aligned. Thank you!!!

Issue #206, opened 2026-05-11
Question about the image size during validation


KostadinovShalon:
Hi, I was wondering if the image size during validation has to be exactly the same as the image size during training or can we define a different image size? The validation method (model.val) has an imgsz argument that seems to not be respected. 

For instance, I have a YOLOX model trained with 640x640 images but I want to test its performance with 1920x1920 (or 1920x1080) images, but when I do imgz=1920 in model.val, I see that the images are processed as in 640x640. I've been tracking down where is the image size set and I saw [here](https://github.com/LibreYOLO/libreyolo/blob/748653f33be954b5202d5ce3475f0d1577cd141a/libreyolo/validation/detection_validator.py#L81) that it's taking the model _native_ input size. 

My question is: is this the expected behaviour? What if I want to train/test with different input images? I feel it should be allowed, and if I'm not wrong, models like YOLOX allow for multi-resolution inputs. However, if it is not allowed or not the expected behaviour, I think it should be documented somewhere. Thanks!

EHxuban11:
Hello @KostadinovShalon , thank you for the detailed report and sorry for the delay.

You are right and was able to reproduce this issue. I fixed it here: https://github.com/LibreYOLO/libreyolo/pull/232

Regaring rectangular inference, this is not planned at the moment.

This kind of report is exactly what we need most at the moment so wanna thank you again.

PR #232, opened 2026-05-22
Honor validation imgsz override

EHxuban11:


EHxuban11:
Fixes https://github.com/LibreYOLO/libreyolo/issues/206

PR #272, opened 2026-05-24
Drop eager dense-mask allocation for polygon segments

### LINKED ISSUE #270: seg masks bug, memory explodes
Commit bd1c8ab (PR #235, "Align RF-DETR fidelity") moved RF-DETR's mask-aware-transform fidelity requirement into the shared dataset annotation parser by routing _yolo_coords_to_rings, the new _yolo_box_to_ring, and _coco_segmentation_to_rings through _attach_dense_mask, which allocates np.zeros((H, W), uint8) per polygon line and wraps the ring in a DenseMaskRing. Because YOLODataset._load_annotations invokes these helpers once per polygon line per image at construction time, opening any non-trivial dataset eagerly allocates a full-resolution uint8 mask per instance before the first batch is produced — for COCO train2017 (118k images, ~8 instances each, ~480x640 px) that's ~280 GB up front, which OOM-killed our loader at a 400 GiB cgroup cap. The change has no opt-out: every load_segments=True consumer pays the cost, including yolo9-seg and ec-seg, which never read dense_mask (the attribute is referenced in exactly one file in the codebase, models/rfdetr/seg_transforms.py) and re-rasterize masks per batch at head resolution in their own loss anyway, so the preloaded full-res rasters are pure waste for them. RF-DETR itself is broken at full-dataset scale by the same mechanism but at least consumes what it allocates; the existing fidelity tests presumably pass because at smoke scale (~50 images) the cost is ~120 MB and invisible. Minimal fix: add attach_dense_masks: bool = False to YOLODataset / COCODataset and thread it through the ring helpers, with only RFDETRTrainer setting it True. Structurally correct fix: do the rasterization lazily inside RF-DETR's _getitem_ (or its collate) so masks live for one worker invocation instead of the lifetime of the dataset.

EHxuban11:
Polygon-sourced rings stored on YOLODataset/COCODataset no longer carry a full-resolution rasterized mask. RF-DETR's crop branch materializes the mask lazily per __getitem__, so crop fidelity is preserved without the dataset-wide allocation that OOM-killed full-COCO loaders (issue #270).

### DIFF
--- a/libreyolo/data/dataset.py
+++ b/libreyolo/data/dataset.py
@@ -32,7 +32,7 @@ def _yolo_coords_to_rings(
     ring = np.array(coords, dtype=np.float32).reshape(-1, 2)
     ring[:, 0] *= width
     ring[:, 1] *= height
-    return _attach_dense_mask([ring], width=width, height=height)
+    return [ring]
 
 
 def _yolo_box_to_ring(cx: float, cy: float, w: float, h: float, width: int, height: int) -> List[np.ndarray]:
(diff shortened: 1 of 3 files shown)


EHxuban11:
bugbot run

EHxuban11:
bugbot run

PR #277, opened 2026-05-25
fix resume LR inconsistency: defer optimizer state and fast-forward s…

imagra93:
## Problem

Resuming training produced an inconsistent learning rate for the first resumed epoch due to two compounding bugs in `BaseTrainer`.

### Bug 1 — Optimizer state silently dropped on resume

The typical call order is `trainer.resume(ckpt)` → `trainer.train()`. `train()` calls `setup()`, which is what creates the optimizer. So when `resume()` ran, `self.optimizer` was always `None`, and the guard:

```python
if self.optimizer is not None and "optimizer" in checkpoint:
```

was never true. The optimizer state dict (momentum buffers, Adam m1/m2, stored LR) was discarded. Every resume effectively started with a cold optimizer.

### Bug 2 — Scheduler LR reset to warmup-start on resume

`_initialize_scheduler_lr()` always called `update_lr(0)`, which returns the near-zero warmup-start LR regardless of the epoch being resumed. The first optimizer step of the resumed epoch therefore ran at the wrong LR before the per-batch update corrected it.

---

## Fix

- **`resume()`**: when the optimizer doesn't exist yet, save the checkpoint state to `_resume_optimizer_state` instead of dropping it.
- **`setup()`**: after creating the optimizer, apply `_resume_optimizer_state` if present (comes after `_initialize_scheduler_lr()` so it takes priority).
- **`_initialize_scheduler_lr()`**: use `start_epoch * steps_per_epoch` as the init iter, fast-forwarding to the correct schedule position on resume.

---

## Tests

Two new tests in `tests/unit/test_trainer_grad_clip.py`:

- **`test_resume_before_setup_defers_optimizer_state`** — verifies the deferred loading path end-to-end, including that the restored LR matches the checkpoint and not the fresh-initialisation value.
- **`test_initialize_scheduler_lr_fastforwards_on_resume`** — verifies that with `start_epoch=5` and `warmup_iters=4`, the scheduler initialises to the post-warmup LR, not the warmup-start value.

### DIFF
--- a/libreyolo/training/trainer.py
+++ b/libreyolo/training/trainer.py
@@ -514,6 +514,17 @@ class BaseTrainer(ABC):
         self.lr_scheduler = self.create_scheduler(self._scheduler_steps_per_epoch())
         self._initialize_scheduler_lr()
 
+        # resume() may be called before setup() when the optimizer doesn't exist
+        # yet. Apply the deferred state now so momentum buffers and LR are restored.
+        if getattr(self, "_resume_optimizer_state", None) is not None:
+            try:
+                self.optimizer.load_state_dict(self._resume_optimizer_state)
+                logger.info("Optimizer state restored from resume checkpoint")
+            except Exception as e:
+                logger.warning(f"Could not load deferred optimizer state: {e}")
+            finally:
+                self._resume_optimizer_state = None
+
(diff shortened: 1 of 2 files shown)


imagra93:
Merged to and continued in [PR-271](https://github.com/LibreYOLO/libreyolo/pull/271)

PR #287, opened 2026-05-27
fix(autobatch): lower default VRAM fraction from 0.70 to 0.60

imagra93:
Lower the default autobatch VRAM target fraction from 0.70 to 0.60. Tested on a lower-spec GPU where 0.70 left too little headroom and caused OOM during training. 0.60 is more conservative and aligns with the fraction used by YOLO.

### DIFF
--- a/libreyolo/models/rfdetr/model.py
+++ b/libreyolo/models/rfdetr/model.py
@@ -145,6 +145,11 @@ class LibreRFDETR(BaseModel):
     multi-scale deformable attention. Segmentation variants add a
     lightweight mask head for instance segmentation.
 
+    autobatch_fraction is lower than the default 0.60 because the probe's
+    fake backward underestimates RF-DETR's real training memory (the loss
+    backward runs through SetCriterion and 6 aux-loss decoder layers), and
+    DDP adds gradient buckets on top.
+
     Args:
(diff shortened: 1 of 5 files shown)


imagra93:
Per-model autobatch_fraction override — RF-DETR multi-GPU training was hitting OOM consistently at epoch 3. The autobatch probe runs a fake backward (t.sum().backward()) which underestimates RF-DETR's real training memory: the actual loss backward goes through SetCriterion with Hungarian matching across 6 aux-loss decoder layers, and DDP gradient buckets are allocated on top. Rather than patching the probe (which would require passing the criterion into autobatch), the simplest fix is a class-level autobatch_fraction = 0.45 on LibreRFDETR. Both autobatch call sites (trainer.py and ddp_spawn.py) read it via getattr(..., _DEFAULT_FRACTION), so other models are unaffected.

scripts/run_sample_train.py — unified reference script for RF-DETR and YOLOv9 on 1 or N GPUs. Exposes --batch (default -1 = AutoBatch, or a fixed integer), --imgsz, --nbs, --resume, and --devices. Intended as a teaching sample, with inline comments explaining the non-obvious options.

imagra93:
@codex review

PR #292, opened 2026-05-28
Fix RT-DETR size-code regex ambiguity and bare GPU index in validator

imagra93:
## Summary

- **RT-DETR size matching** — tighten the fallback regex to require a
  word boundary after the size code, preventing `"-l"` from matching
  inside `"-large"` (and similar collisions between size tokens).
- **Validator device** — normalise bare GPU index strings (`"0"`, `"0,1"`)
  to `cuda:0` before passing to `torch.device`, consistent with how the
  trainer already handles this.

### DIFF
--- a/libreyolo/models/base/inference.py
+++ b/libreyolo/models/base/inference.py
@@ -212,7 +212,8 @@ class InferenceRunner:
         target = torch.device(device_str)
         if target != self.model.device:
             self.model.device = target
-            self.model.model.to(target)
+            if hasattr(self.model.model, "to"):
+                self.model.model.to(target)
 
     def _process_in_batches(
         self,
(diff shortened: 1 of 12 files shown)


imagra93:
## Updated Summary

- **RT-DETR size matching** — tighten the fallback regex to require a
  word boundary after the size code, preventing `"-l"` from matching
  inside `"-large"` (and similar collisions between size tokens).
- **Validator device** — normalise bare GPU index strings (`"0"`, `"0,1"`)
  to `cuda:0` before passing to `torch.device`, consistent with how the
  trainer already handles this.
- **RF-DETR size auto-detection** — `LibreRFDETR` now infers the correct
  size variant from the weight filename when `size` is not specified
  explicitly. Handles both `LibreRFDETR*.pt` names and upstream
  `rf-detr-{nano,small,medium,large}*.pth` names, fixing a crash where
  loading `rf-detr-large-2026.pth` without `size="l"` would build a
  3-layer model against a 4-layer checkpoint.

EHxuban11:
We are trying to avoid filename parsing. The LibreYOLO size detection is metadata-based, documented here: docs/checkpoint_schema.md

PR #294, opened 2026-05-28
Fix/rfdetr seg imgsz override and ddp patch size

imagra93:
Problem

Two compounding bugs caused a crash when training LibreRFDETR seg with a user-supplied imgsz under DDP (e.g. size="l", imgsz=624, 4 GPUs), plus a separate DDP reducer error.

Bug 1 — imgsz silently overridden (model.py)

train_kwargs.update({"imgsz": self.input_size}) was called after train_kwargs = dict(kwargs), so imgsz=624 was already present but got clobbered by the model default (504 for l-seg). Training continued at the wrong resolution.

Bug 2 — Wrong multi-scale divisor after DDP wrapping (trainer.py)

_multi_scale_scales() is called per-batch during training, by which point self.model has been wrapped in DistributedDataParallel. DDP does not proxy custom attributes, so getattr(ddp_model, "patch_size", 16) always fell back to 16 (correct: 12) and num_windows to 4 (correct: 2). This gave divisor 64 instead of 24, generating scale 512 — which fails the backbone's assertion 512 % 24 == 0, crashing the first training step.

Bug 3 — DDP "marked ready twice" crash (trainer.py)

The seg head is invoked in two branches per forward pass (encoder queries and decoder queries), so the same parameters accumulate gradients from both branches. With find_unused_parameters=True this causes DDP's per-parameter hooks to fire twice → RuntimeError: Expected to mark a variable ready only once.

Fix

model.py: replace train_kwargs.update({"imgsz": self.input_size}) with if train_kwargs.get("imgsz") is None: train_kwargs["imgsz"] = self.input_size, so an explicit user value (including imgsz=0) is never overridden. setdefault alone is insufficient because it would let imgsz=None pass through to the trainer.
trainer.py: call unwrap_model(self.model) before reading patch_size / num_windows in _multi_scale_scales().
trainer.py: override _ddp_find_unused_parameters() → False and _ddp_static_graph() → True in RFDETRTrainer. static_graph=True locks the DDP reducer graph after the first iteration, eliminating the double-ready error without requiring find_unused_parameters.

### DIFF
--- a/libreyolo/cli/config.py
+++ b/libreyolo/cli/config.py
@@ -377,6 +377,9 @@ def _build_rfdetr_train_kwargs(
             kwargs[target_name] = params[cli_name]
 
     provided = user_provided or set()
+    if "imgsz" in provided and params.get("imgsz") is not None:
+        kwargs["imgsz"] = params["imgsz"]
+
     if "patience" in params:
         kwargs["early_stopping"] = params["patience"] > 0
         kwargs["early_stopping_patience"] = params["patience"]
--- a/libreyolo/models/rfdetr/model.py
+++ b/libreyolo/models/rfdetr/model.py
@@ -180,7 +180,6 @@ class LibreRFDETR(BaseModel):
 
     # CLI parameters intentionally ignored by native RF-DETR training.
     UNSUPPORTED_TRAIN_PARAMS: ClassVar[set[str]] = {
-        "imgsz",
         "mosaic",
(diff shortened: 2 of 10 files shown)


EHxuban11:
My agent flags serveral changes that are unrelated with the PR description and some of them affect the whole library:

RT-DETR filename parsing, changing base trainer device sync for all trainers, global pytest default changes.

PR #313, opened 2026-05-31
312 qa for the v120 release

### LINKED ISSUE #312: QA for the v1.2.0 release
Very important:
- Remove YOLO9 E2E supported task segment in SUPPORTED_TASKS=("detect","segment")
- Pump package number to v1.2.0
- RF-DETR x/xx seg weights unavailable, investigate if I can redistribute them
- RF-DETR seg models dont have readmes and licenses in hf
- RF-DETR forces to use size"" which is not good
- RF-DETR train() mishandles project/name/exist_ok and treats resume=True as a literal path
- There is no .yaml files inside the wheel or the sdist, fix that in pyprojects.toml or MANIFEST.in (idk where rn). Also add teh .txt files!!
- Increase the dependency lower bounds because an LLM detected for example that torch>=1.13.0 is too low for torch.amp.GradScale for instance. torchvision>=0.11.0 is too low for transforms.v2 / tv_tensors, Pillow>=8.0.0 is too low for Image.Resampling, transformers>=4.40.0 is too low for RF-DETR’s current DINOv2 imports.
- Update readme
- 

Improvements:
- [ ] plot(), save()

Other secondary things:
-  Fix release workflow/docs mismatch RELEASING.md implies publish workflow runs CI, but publish only builds/publishes. Make docs explicit: tests must pass before tagging, or make publish depend on checks.
-  Fix libreyolo val --json --quiet It returns valid metrics eventually, but stdout is polluted by COCOeval text, so it is not parseable JSON. Also --no-use-coco-eval is currently ignored.
-  Fix the Windows RF-DETR reload test harness, flagship nightly fails on Windows due to \U path escaping.

EHxuban11:


EHxuban11:
@claude review

EHxuban11:
@codex review

EHxuban11:
@cursor review

Issue #291, opened 2026-05-28
[Bug] Training error occurs when using the latest compiled code for RF-DETR


wuhaoyu010:
**Description**
I compiled the latest code from the repository and attempted to train the RF-DETR model, but encountered an error.
**Screenshot**

<img width="1445" height="201" alt="Image" src="https://github.com/user-attachments/assets/55311d64-a22b-4ccd-b2e5-4f4827dc6bac" />

**Training Code**
```python
from libreyolo import LibreRFDETR
def main():
    model = LibreRFDETR("weights/rf-detr-large-2026.pth")
    result = model.train(
        data="datasets",
        epochs=100,
        batch=16,
        device="4,5",
        imgsz=640,
        lr=1e-4,
        output_dir="runs/train/rfdetr_exp",
    )
    print(f"Training completed: {result}")
if __name__ == "__main__":
    main()
```
**Additional Context**
I have tried all three pre-trained models available in `LibreYOLO/LibreRFDETR`, but the same error occurs with all of them. Could you please help me identify what might be causing this issue? Thanks!

<img width="1552" height="555" alt="Image" src="https://github.com/user-attachments/assets/1b869057-eb91-4126-93d4-00a231f54efc" />

imagra93:
Hi, thanks for the detailed report!

We are are working in this bug — when loading a weight file directly (e.g. rf-detr-large-2026.pth) without specifying size=, the model defaults to size="s" (3 decoder layers) instead of the correct size="l" (4 decoder layers), causing the unexpected-keys error on transformer.decoder.layers.3.*. The same mismatch applies to the medium and nano checkpoints.

A fix will be merged to dev today. After the update, LibreRFDETR("weights/rf-detr-large-2026.pth") will automatically detect the correct size from the filename — no code changes needed on your end.

In the meantime, you can work around it by passing the size explicitly:

```
model = LibreRFDETR("weights/rf-detr-large-2026.pth", size="l")
model = LibreRFDETR("weights/rf-detr-medium.pth", size="m")
model = LibreRFDETR("weights/rf-detr-nano.pth", size="n")
```

wuhaoyu010:
Hello，你的来信已收到

wuhaoyu010:
Thanks for the reply. However, after applying the changes, I am still encountering the following error.
**Training Code**
```python
from libreyolo import LibreRFDETR
def main():
[...]

PR #323, opened 2026-06-03
fix(seg): keep RF-DETR mask loss graph connected

imagra93:
## Problem

RF-DETR segmentation training with `static_graph=True` can crash when a rank sees a batch with zero Hungarian matches, such as an all-background batch.

In the factored mask-head branch, `loss_masks` returned a disconnected empty tensor for the no-match case. That tensor has no `grad_fn`, so the mask-head tensors do not participate in backward on that rank, which violates DDP's static-graph requirement.

Background/unlabeled images are valid negative samples and should remain in the training dataset.

Addresses the DDP unlabeled-image crash in #291.

## Solution

Return zero mask losses routed through the mask-head tensors instead of a disconnected empty tensor:

```python
zero = spatial_features.sum() * 0.0 + query_features.sum() * 0.0 + bias.sum() * 0.0
return {"loss_mask_ce": zero, "loss_mask_dice": zero}
```

The loss value is still zero, but `spatial_features`, `query_features`, and `bias` remain in the backward graph and receive zero gradients. This keeps DDP's reducer stable without dropping background images.

A regression test covers the no-match mask-loss path and asserts that backward keeps all mask-head tensors connected with zero gradients.

### DIFF
--- a/libreyolo/data/dataset.py
+++ b/libreyolo/data/dataset.py
@@ -160,7 +160,6 @@ class YOLODataset(Dataset):
         img_files: List[Path] | None = None,
         label_files: List[Path] | None = None,
         load_segments: bool = False,
-        filter_empty_annotations: bool = False,
     ):
         """
         Initialize YOLO dataset.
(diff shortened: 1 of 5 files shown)


imagra93:
@codex review

EHxuban11:
I'm concerned that this PR is dropping the unlabelled images basically (often called background images)d, but background images are useful in order to avoid the model to detect false positives when there isn't an object in the image.

imagra93:
Correct: I keep your unlabeled images. The DDP crash is gone.

A selective filter (e.g. max_unlabeled_ratio=0.X) could be a useful future option for datasets where the unlabeled fraction is so large it dominates the sampler and starves the model of positive examples. But the right threshold is dataset-dependent and we don't have enough testing to pick a sensible default yet, so we're leaving that for a follow-up if there's demand.

EHxuban11:
Thank you @imagra93 for your contribution. This is mergeable as it is but i'll add a small test first.

PR #386, opened 2026-06-11
384 support in memory batch processing via lists in predict

### LINKED ISSUE #384: Support in-memory batch processing via Lists in predict()
First of all, thanks for LibreYOLO!

Currently predict() does not support true parallel batching unless the source is a directory with files.

If I try to pass a list of cv images it fails with "TypeError: Unsupported image type: list. Supported types: str, Path, PIL.Image, np.ndarray, torch.Tensor, bytes, BytesIO"

Currently the TensorRTBackend already implements a _process_in_batches that could be used for this purpose.

The solution can be something simple, like so:
`        if isinstance(source, list):
            return self._process_in_batches(source...)`

Please implement this so we can use batch inference on in-memory images.

This is trivial when using Tensor backend, but when using pytorch backend it is harder. Please implement with pytorch as well.

Thanks!

EHxuban11:
## Summary by CodeRabbit

* **New Features**
  * Batch inference accepts in-memory image lists/tuples and generates deterministic indexed output names to avoid overwrites.
  * Backends now detect and respect dynamic-batching capability and route list inputs to batched processing when supported.

* **Bug Fixes**
  * NCNN rejects unsupported batched inputs instead of returning incorrect results.
  * Consistent per-image extraction from batched model outputs.

* **Documentation**
  * CLI batch help text clarified.

* **Tests**
  * Large unit and e2e test suite added for list-batch parity and backend behaviors.

### DIFF
--- a/libreyolo/backends/base.py
+++ b/libreyolo/backends/base.py
@@ -1699,7 +1699,14 @@ class BaseBackend(ABC):
         of ``batch`` images runs as a single forward pass; otherwise images
         run sequentially.
         """
-        if batch > 1 and self._supports_batched_inference():
+        use_batched = (
+            batch > 1
+            and self._supports_batched_inference()
+            # Latched by _predict_batch after a runtime rejects a stacked
+            # blob, so a long list does not retry (and warn) once per chunk.
+            and not getattr(self, "_batched_inference_failed", False)
+        )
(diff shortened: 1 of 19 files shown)


EHxuban11:
@codex review

PR #398, opened 2026-06-14
Feat/librelabel design

EHxuban11:
## Release Notes

* **New Features**
  * Added a `label` CLI subcommand to start a local web-based dataset annotation server (host/port, device selection, browser/assist toggles, JSON/quiet/verbose output, LAN sharing).
  * Includes offline AI-assisted suggestions, bulk autolabeling with live progress, thumbnails, dataset insights, and interactive box/polygon editing.
  * Adds click-to-mask segmentation, plus dataset “error radar” insights, and embedding-based visualization; includes train-in-the-loop “boost” using accepted labels.
* **Documentation**
  * Added an ADR describing LibreLabel clean-room provenance rules.
* **Tests**
  * Added unit tests for label round-trips, sanitization, assist mapping, radar/quality checks, and safeguard behaviors.


EHxuban11:
## Addressed in `7cd1d3d9` (and `d754575b`)

Thanks @chatgpt-codex-connector and @coderabbitai — worked through the full review.

**P1**
- **Pose/keypoint data loss** → keypoint/pose datasets (`kpt_shape`) are now view-only, and any label file with unsupported (keypoint/malformed) rows is read-only, so a save can never drop keypoints or corrupt OBB. (`dataset.py` + `labelio.has_unsupported_rows`, with tests)
- **CSRF on label writes** → cross-origin POSTs are now rejected before any write (`Origin` vs `Host` check), so a foreign site can't overwrite `labels/*.txt` via the localhost server. (`server.py._same_origin`)

**P2 — fixed**
- AI-assist POSTs rejected under `--no-assist`; class-key/picker updates a selected **polygon** too; dirty state preserved if you edit during an in-flight save; assist model list limited to **box-producing** tasks; Locate Anything gated on the VLM extra (`decord`); SAM normalizes numeric devices (`0`→`cuda:0`) and loads **strictly offline** (cache-checked, `HF_HUB_OFFLINE`, respects `HF_HOME`/`HF_HUB_CACHE`); degenerate polygons dropped on save; reviewed-**empty** files skipped by auto-label; reject/clear + re-run no longer leave phantom "to review" rows; **exact same-path** train/val leakage now surfaced.

**CodeRabbit** — `strict=True` on the parallel zips, SAM point-`None` guard, `verbose`-arg comment.

**Left as-is (deliberate):** *detector assist auto-downloads weights.* This is intentional — the detector is the user's **own** model, fetched once from the project's HF mirror exactly like every other `libreyolo` model load, then cached. The "local/offline" contract means no cloud labeling service, no account, no telemetry, and on-device inference — not "never fetch the chosen weights". Forcing local-only would break first use of the headline auto-label feature. Happy to add a `--local-models-only` strict mode if preferred.

All green locally: ruff, embedded-JS parse, 22 unit tests.

@coderabbitai review

EHxuban11:
@codex review — all 14 actionable findings from the last pass are addressed in 7cd1d3d9 (details in the comment above). Please take another look.

EHxuban11:
Second review round addressed in 684e4c4. All 8 items:

1. **Quarantine basename collision** — `resolve_duplicates` now prefixes the quarantine destination with the source index, so two images that share a basename across splits no longer overwrite each other in the quarantine dir.
2. **`write_label` read-only recheck** — re-runs `has_unsupported_rows` immediately before writing, so a file that gained a keypoint/pose row after open can't be clobbered.
3. **`images_*` substring ambiguity** — `_check_writable` now anchors on an exact `images` / `images_*` path *segment*, not any ancestor that merely contains the substring "images".
4. **Autolabel session guard (P1)** — `autolabel_dataset` takes a `current()` identity callback and stops populating `pending` the moment the active project switches mid-stream; `server` wires it in.
5. **`load()` after auto-save** — re-derives the active image post-save so a deleted/switched image can't be re-shown.
6. **Dense labels view-only** — `masks_dir` / `depths_dir` / `depth` datasets are now view-only (we don't round-trip dense labels), same contract as keypoints/pose.
7. **Leakage panel** — driven off the leakage payload, not only `duplicate_groups`.
8. **Read-only canvas** — pointerdown is select-only and Delete/setActive/prelabel are guarded on read-only datasets.

+3 regression tests (unsupported-row write refusal, images-substring ancestor stays read-only, mask dataset view-only). 25 pass, ruff clean.

@codex review

EHxuban11:
Third review round — addressed **all** open findings from Codex, CodeRabbit, and Cursor in 2189295.

**Read-only / dense-label contract**
- Task-only `task: depth|classify|pose|obb` yamls are now view-only (a depth yaml may omit `depths_dir`, classify uses no `.txt`, OBB 9-field rows are oriented rects we'd corrupt). *(Codex `dataset.py:95`, `labelio.py:138`)*
- `_check_writable` now keys off *starts-with* `images` (matching `img2label_paths`), so `my_images` stays writable while `images_2026`/nested `images` stay read-only. *(Codex `dataset.py:115`, CodeRabbit)*
- `write_label` rejects tombstoned/quarantined ids. *(Codex `dataset.py:411`, CodeRabbit `dataset.py:374`)*
- Canvas: polygon vertex insert (dblclick), Tighten (T), and ghost acceptance are blocked on read-only; accepting an **unmatched** ghost now applies the active palette class as the UI promises. *(Codex `page.py:1385/1579/1122/1106`)*

**Label parsing**
- Fractional class token `0.5` → unsupported/read-only (not coerced to `0`); `OverflowError` caught. *(Codex `labelio.py:123`, CodeRabbit `:160`)*
- Collinear polygons dropped by shoelace **area** (catches diagonal degenerates). *(Codex `labelio.py:196`)*

**Assist / session safety**
- `enabled` enforced before any inference (closes the `--no-assist` bypass); `_deleted` tombstones skipped; active session re-checked right before storing pending. *(CodeRabbit `assist.py:165/266`, Codex `:282`, Cursor)*
- LocateAnything (non-commercial) is now **opt-in only** via `LIBRELABEL_ENABLE_LOCATE=1` — never advertised/loaded by default. It is an existing optional `LibreVLM` integration, not vendored code. *(CodeRabbit `assist.py:222`)*

**Boost** — one locked session snapshot for the whole start; JSON-encoded YAML names/paths; process-wide torch thread count saved & restored. *(CodeRabbit `boost.py:64/140/191`)*

**Other** — purge deletes label-then-image (recoverable partial state); embed cache re-checked & published under the lock; radar class-slips gated at `miss_conf`; `projects` tolerates malformed `last_opened`; `sam` checks the real `weights/LibreSAM<size>/` snapshot and **refuses-if-not-cached** instead of mutating process-wide `HF_HUB_OFFLINE`; server/CLI publish the actually-bound port (`port=0`) and a machine-parseable `lan_url`; ambiguous `×`/`–` → HTML entities. *(CodeRabbit `dataset.py:371`, `embed.py:99`, `radar.py:99`, `projects.py:79`, `sam.py:65/84`, `server.py:588`, `label.py:90`, `page.py:776`)*

Note on `sam.py:84`: the proper `local_files_only=True` plumbing lives in `libreyolo/models/sam/base.py`, which belongs to the LibreSAM PR (not this diff) — so I removed the env mutation here entirely and gate on a verified local snapshot instead, which is thread-safe and keeps the offline contract.

Validation: **31 unit tests pass** (+6 regressions), ruff clean, RUF001 clean, embedded JS parses, bound-port smoke test green.

@codex review

EHxuban11:
Fourth review round — all 10 Codex findings addressed in 45977ee.

**Session / epoch races (P1)**
- Single-image prelabel now snapshots the session and only stores suggestions (else 409) if the project didn't switch mid-inference. *(`server.py:423`)*
- `/api/insights/fix` is epoch-guarded like label saves; `page.js` sends `DS.epoch` with the Fix request, so a duplicate-fix racing a switch can't quarantine same-id images in the new project. *(`server.py:356`)*

**LocateAnything opt-in (P2)**
- Opt-in now requires an explicit affirmative value (`1/true/yes/on`); `0`/`false` no longer slip through the truthiness test. *(`assist.py:221`)*
- `predict_locate` fails closed on the opt-in, so the hidden `engine=locate` endpoint can't bypass it. *(`assist.py:248`)*

**Offline contract (P2)**
- Assist refuses to load a model whose weights aren't already local (clear hint) instead of letting the factory auto-download — a local labeller shouldn't silently hit the network on first run. *(`assist.py:152`)*

**Label integrity (P2)**
- `parse_annotations` uses the integer-class contract: a `0.5` row is skipped, not coerced to class `0`, so `stats()`/Radar/Boost (which don't check `editable`) never count/train it as `0`. *(`labelio.py:123`)*
- `resolve_duplicates` keeps a **labelled** survivor (then train, then lowest id) so Fix can't turn the only labelled copy unlabelled, and refuses inline YAML image-list splits too (not just `.txt`). *(`dataset.py:342/347`)*

**UX (P3)** — keyboard accept-all applies the active palette class to unmatched ghosts (same fallback as single accept). *(`page.py:1125`)*

**CLI (P2)** — `label --json` includes `lan_url` in the eager-dataset response, not just home mode. *(`label.py:105`)*

Validation: **36 unit tests** (+5 regressions), ruff clean, JS parses.

Note on `assist.py:152`: I chose refuse-at-run-time (with a fetch hint) over hiding all models from the UI, so the assist panel still lists models but never triggers a surprise download — making the fetch an explicit user action, as suggested.

@codex review

EHxuban11:
Fifth review round — all 7 Codex findings addressed in cb12546.

**Offline contract (P2)** — the fail-closed local-weight check now covers *every* model-backed action, not just assist. Factored into `assist.weight_is_local()` and applied before model construction in:
- **embed** (`embed.py:51`) — opening the Map no longer triggers a silent download.
- **Boost** (`boost.py:207`) — starting Boost no longer downloads the base checkpoint.

**Session race (P2)** — `server.py:437`: the prelabel pending write is now atomic with the project switch. `_LabelState.store_pending()` does the session-identity check + `set_pending()` under the same `_lock` that `open_project()` uses to clear pending and rebind the session, so a switch can't interleave and leave stale suggestions under a new project's numeric id.

**Label integrity**
- `parse_label_text` (the box-only format oracle) now uses the integer-class contract — `0.5 ...` is non-box/unsupported, matching `parse_annotations` and the loader. *(`labelio.py:52`)*
- `save()` uses the server's returned `count`: if a degenerate shape was sanitized away, it surfaces "N invalid shape(s) dropped" instead of reporting a clean save. *(`page.py:907`)*

**UX (P3)** — single-image prelabel marks the row `suggested` (like bulk autolabel) so manual prelabels are findable via the Review filter. *(`page.py:1037`)*

**Docs (P3)** — renumbered the LibreLabel ADR `0006 → 0007` (0006 was already the depth-task contract). *(`docs/adr`)*

Validation: **39 unit tests** (+3 regressions), ruff clean, JS parses.

@codex review

EHxuban11:
Sixth review round — all 7 Codex findings addressed in 3f0f2be.

**Session atomicity (P2)** — `assist.py:343`: bulk auto-label now publishes each image's suggestions through a state-level `store(idx, sugg)` callback that checks session identity and writes pending under `_LabelState._lock` (the same lock `open_project()` holds to clear pending + rebind), so a bulk run can't leak stale boxes into a switched-to project's numeric ids.

**Systemic model-load failures abort instead of "0 boxes"**
- `assist.py:334`: `autolabel_dataset` pre-flights the model once before the loop, so a missing-weight / opt-in / disabled error propagates and the stream emits it.
- `radar.py:155`: `scan_dataset` propagates `RuntimeError` (model can't run at all) instead of swallowing it per-image into "no labelled images to audit".

**Non-finite labels (P2)** — `labelio.py:158`: `nan`/`inf` coordinates are unsupported → the file stays read-only; `parse_annotations` skips them (no NaN into stats/Radar/JSON); `parse_label_text` and both sanitizers drop them so they never reach a label file.

**page.js**
- `save()` (`page.py:905`): when the server sanitizes away a degenerate shape (`count < sent`), the edit stays dirty/failed (warn + `return false`) so close/nav can't silently lose it. *(P2)*
- single-image prelabel returning no suggestions clears the stale Review row instead of leaving a phantom `suggested` entry. *(`page.py:1044`, P3)*
- `resetClientState()` resets `listFilter` to `"all"` on project switch, so a new project can't open with a Review/To-do filter hiding every image. *(`page.py:1815`, P3)*

Validation: **41 unit tests** (+2 regressions), ruff clean, JS parses.

@codex review

EHxuban11:
Seventh review round — all 8 Codex findings addressed in 616380a.

**Shared-server security (P2)** — `server.py:333`: `/api/projects/open` & `/forget` are now restricted to the host's own loopback client when bound with `--share` (0.0.0.0). Switching projects rebinds the global session for everyone, so a LAN teammate can no longer repoint the host at an arbitrary local `data.yaml` and read/write its labels.

**Non-finite labels — the real fix (P2)** — `labelio.py:83`: finiteness is now checked on the **raw** float *before* `_clamp01()` in both sanitizers. `Infinity`/`-Infinity` (which `json.loads` accepts) was being clamped to a finite `1.0`/`0.0` edge and kept; now dropped.

**Systemic model-load failures (P2)** — `embed.py:135`: the embedding Map joins Radar/auto-label and pre-flights the model, so missing local weights surface the actionable error instead of an empty "done" with no points.

**save / navigation integrity (page.js)**
- `save()` (`page.py:895`): compares the server count against **every** canvas shape, so a box clipped to zero area (filtered client-side) is also a failed/dirty save, not a silent vanish on reload.
- `load()` (`page.py:848`): after a successful save, stays put if the image is still dirty (edits during the POST) instead of discarding them.
- `runBoost()` (`page.py:1643`): aborts if the current save fails, so Boost can't train/report on stale labels.
- auto-label stream (`page.py:1176`): an NDJSON `error` sets a failed flag and shows the error instead of falling through to "Done / 0 boxes".
- `setRowStatus()` (`page.py:696`): a hidden row that newly matches the active filter is brought into the sidebar (debounced re-render).

Validation: **42 unit tests** (+1 regression), ruff clean, JS parses.

@codex review

EHxuban11:
Eighth review round — all 4 Codex findings addressed in 6485469.

**Shared-server admin gating (P2)**
- `_local_admin` (`server.py:411`): only a *loopback bind* (127.0.0.1/localhost) is inherently local-only. A concrete NIC bind (`host=192.168.x.y`) is LAN-reachable but was treated as all-admin — now any non-loopback bind requires a loopback *client*.
- `/api/insights/fix` (`server.py:315`) is gated the same way: it moves/unlinks image+label files via `resolve_duplicates(purge)`, so a shared client can't quarantine or hard-delete dataset entries — host-admin only.

**Navigation / polling (page.js, P2)**
- `pollBoost` (`page.py:1679`): guards on the project epoch (`DS.epoch`), not `loadSeq`. `loadSeq` bumps on every image navigation, so moving images mid-boost killed the poll and the UI never showed completion / selected the boosted model; the epoch only changes on a real project switch.
- `backToHome` (`page.py:1801`): re-checks `dirty` after `save()` (like `load()`), so clicking Home while an in-flight save left edits dirty can't open the home screen with unsaved edits that `resetClientState()` would then drop.

Validation: **43 unit tests** (+1 regression), ruff clean, JS parses.

@codex review

EHxuban11:
Ninth review round — all 6 Codex findings addressed in 63db3d8.

**Shared-server admin gating, refined (P2)**
- `_local_admin` (`server.py:419`): only a *wildcard* bind (`0.0.0.0`/`::` — what `--share` uses) requires a loopback client. A concrete NIC bind (`host=192.168.x.y`) forces the host's own browser to connect via that address, so the round-8 loopback-client check made admin unusable for the host — now allowed on explicit NIC binds (an advanced opt-in exposure), while the common `--share` path stays gated.
- `GET /api/projects` (`server.py:202`): redacts the registry (host-local dataset paths) for non-admin LAN clients, who can only label the already-open project.

**Navigation / dirty races (page.js, P2)**
- `reviewFinding` (`page.py:1560`): overlays Radar findings only if `load()` actually landed on the target image (`idx===id`), so a refused/superseded load can't draw another image's findings onto the open one.
- `runBoost` (`page.py:1665`): re-checks `dirty` after `save()` — edits landed during the POST won't start Boost on stale labels.
- `carryForward` (`page.py:978`): captures the load generation + source idx before the async fetch and bails if either changed, so the previous image's labels can't paste onto whatever image is now open.
- `fixDuplicate` (`page.py:824`): saves (and re-checks dirty) before `/api/insights/fix`, so the server can't quarantine the dirty copy before its only labelled edits are written.

Validation: **43 unit tests**, ruff clean, JS parses.

@codex review

EHxuban11:
Tenth review round — all 7 Codex findings addressed in 065753e.

**Label integrity (P2)**
- `dataset.py:449`: a file with an integer class outside `[0, nc)` is now read-only (`read_label` editable=False + `write_label` rejects), so an unrelated save can't sanitize that annotation away. New `labelio.has_out_of_range_rows`.
- `boost.py:107`: `_gather` skips read-only (keypoint/OBB/out-of-range) files — their parsed boxes are a partial view, so training/agreement on the subset would skew.

**Shared-server path leak (P2)** — `server.py:225`: `GET /api/dataset` redacts `root`/`yaml` (host-local paths) for non-admin LAN clients, keeping names/count/writable needed to label.

**LAN teammate UX (P2)** — `server.py:207`/`page.py`: a non-admin teammate who clicks Home is no longer stranded — Home offers "Resume current project" whenever a session is open (resume re-enters the live session, no admin needed).

**Navigation / dirty races (page.js, P2)**
- `fixDuplicate` (`page.py:834`): if Fix quarantines the open image, clear dirty and navigate to the survivor (`d.kept`)/a live image instead of leaving a tombstoned canvas whose next save is rejected.
- `runRadar` (`page.py:1529`): re-checks dirty after `save()` (like Boost).
- `carryForward` (`page.py:985`): re-checks emptiness after the async fetch, so a label drawn during it isn't merged with the carried-forward ones.

Validation: **44 unit tests** (+2 regressions), ruff clean, JS parses.

@codex review

EHxuban11:
Eleventh review round — all 9 Codex findings addressed in 3abdf61.

**Concurrent-edit safety on `--share` (P2)** — `server.py:126`: optimistic concurrency for `/api/label`. GET returns a `rev` (label-file mtime); POST sends the loaded rev and is rejected (409) if the file changed since, so two LAN teammates editing the same image can't silently clobber each other. The save adopts the returned rev so a user's own next save doesn't self-conflict.

**Label integrity (P2)**
- `dataset.py:436`: a file is editable only if the dataset is writable AND no row would be altered by a save — now also out-of-`[0,1]` coordinates (`has_out_of_bounds_coords`) and, via the writable gate, dense-task rows (e.g. a `kpt_shape:[17,2]` pose row the generic polygon branch would otherwise accept — `labelio.py:172`).
- `radar.py:145`: skip non-editable (partial-view) files — no false misses/class-slips the user can't fix.

**Navigation / race / read-only guards (page.js, P2)**
- `load()` (`page.py:879`): canvas stays inert (`imgOk=false, editable=false`) until metadata loads, with an error state if `/api/label` fails — a failed fetch can't leave edits saving to a new id with the old image's dimensions.
- `openInsights` (`page.py:752`): bound to `DS.epoch`, drops stale results after a project switch.
- `carryForward` (`page.py:998`): refuses a read-only source label (`editable===false`).
- `segmentAt` (`page.py:1111`): adds the `DS.writable` guard so Smart Segment can't add an unsavable polygon.
- `autolabelAll` (`page.py:1188`): re-checks dirty after `save()`.

Validation: **47 unit tests** (+2 regressions), ruff clean, JS parses.

@codex review

EHxuban11:
Twelfth round + a **preemptive self-audit** — addressed in 468425c.

I ran a parallel multi-agent audit of the whole codebase against every recurring finding-class and fixed everything it confirmed, folded together with this round's review.

**P1 (round 12) — `dataset.py:450`:** revision tokens are now serialized as **strings**. `st_mtime_ns` (~1e18) exceeds JS `Number.MAX_SAFE_INTEGER`, so the numeric token was rounded and *every* save would have 409'd. Good catch — this was a regression from round 11's concurrency feature.

**Round 12**
- `dataset.py:432`: `read_label` no-file branch returns the session writable flag (read-only sessions stay inert). `dataset.py:307`: geometry-lint skips non-editable files. `dataset.py:103`: 4-point OBB-shaped rows are read-only unless `task: segment` (the taskless-OBB case — a 4-point row is byte-identical to an OBB per the schema, so it's protected conservatively).
- `server.py:164`: thumbnail cache keyed by image path (no cross-project leak).
- `page.py:1210`: the dataset-wide auto-label/Radar/embedding streams guard on `DS.epoch`, not `loadSeq`, so A/D navigation can't abort them. `page.py:1179`: `clearReviewState` keeps a present-but-empty file as `empty` (reviewed background).

**Preemptive audit (12 verified sites, would have been ~5 future rounds)**
- **Data loss:** `openProject` now saves/guards before switching projects (it was silently discarding unsaved edits).
- **Shared-server:** gated the heavy server-global compute streams (`/api/boost`, `/api/assist/autolabel|radar`, `/api/embeddings`) behind `_local_admin`; redacted host paths from non-admin error bodies and `/api/dataset` `reason`.
- **Integrity:** zero-area/collinear polygons on disk are read-only (`has_degenerate_polygon`).
- **Nav races:** `renderStats` + the `fixDuplicate` continuation re-check `DS.epoch`.
- **Systemic failures:** LocateAnything VLM pre-flighted; `boost._agreement` re-raises instead of publishing a fabricated 0% as "done".

Validation: **50 unit tests** (+5 regressions), ruff clean, JS parses.

@codex review

EHxuban11:
Thirteenth round — all 7 Codex findings addressed in b236be0. These were mostly second-order edges of the round-11/12/audit changes.

**Shared-server**
- `page.py:1216`: `autolabelAll` now checks `r.ok` before streaming — now that the full-dataset streams are admin-gated, a JSON 403 isn't consumed as NDJSON (which read as a "successful 0-box run").
- `server.py:327`: `/api/thumb` 415 error redacted for non-admin (PIL errors embed the filename).
- `server.py:331`: project epoch appended to `/api/thumb` URLs so the browser cache can't serve a previous project's thumbnail for a reused numeric id (server cache was already path-keyed; this is the client side).

**Label integrity** — `dataset.py:446`: a zero-width/height box on disk (`has_zero_area_box`) keeps the file read-only, same class as the degenerate-polygon guard — the sanitizer would otherwise drop the row on any save.

**Concurrency / duplicate-fix**
- `server.py:269`: annotations + revision are read under the save lock (`read_label_with_rev`), so a save landing between can't hand a stale client old annotations with a new rev token.
- `dataset.py:385`: purge deletes the **image first** — a failed image unlink leaves the labelled pair intact instead of orphaning the deleted label.
- `dataset.py:372`: refuse pruning a recursive directory split (`train: .`) whose rglob would re-discover the quarantine folder.

Validation: **51 unit tests** (+2 regressions), ruff clean, JS parses.

@codex review

EHxuban11:
@codex review

EHxuban11:
@codex review

EHxuban11:
@codex review

PR #404, opened 2026-06-16
399 add tensor rt to every model

### LINKED ISSUE #399: add tensor rt to every model

EHxuban11:
## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Improved PICODET and DAMO-YOLO post-processing with multi-label confidence selection, score-based candidate caps before NMS, and extended class-aware NMS for YOLOX/PICODET.
  * Reduced DAMO-YOLO retained NMS candidates.
  * Fixed ONNX/TensorRT bilinear sampling by using a manual gather-based implementation during export.
  * Improved numerical stability by replacing `inf` invalid-anchor sentinels with a safe finite value for TensorRT/FP16.
* **New Features**
  * Enhanced TensorRT fast FP16 export with ViT-like backbone detection and backbone layer precision pinning.

### DIFF
--- a/libreyolo/backends/base.py
+++ b/libreyolo/backends/base.py
@@ -700,6 +700,10 @@ class BaseBackend(ABC):
         boxes = boxes[box_indices].copy()
         max_scores = scores[box_indices, class_ids]
 
+        if max_scores.shape[0] > _YOLO9_MAX_NMS_CANDIDATES:
+            keep = np.argpartition(-max_scores, _YOLO9_MAX_NMS_CANDIDATES - 1)[:_YOLO9_MAX_NMS_CANDIDATES]
+            boxes, max_scores, class_ids = boxes[keep], max_scores[keep], class_ids[keep]
+
         scale_x = orig_w / effective_imgsz
         scale_y = orig_h / effective_imgsz
         boxes[:, [0, 2]] *= scale_x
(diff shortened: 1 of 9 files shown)


EHxuban11:
Addressed all review feedback across the latest commits (HEAD \`75235c5f\`):

- **ViT FP32 pin** now applies to INT8 builds too (FP16 flag is set for \`half or int8\`).
- **PicoDet cap** is now **per-FPN-level** with **ceil** feature sizes, so picodet-m (416 → 7×7 P6, N=3598) actually engages per-level instead of falling back to global. Parity preserved (s/m/l = 0.304/0.379/0.441 = native).
- **\`precision_str\`** only reports "FP32 ViT backbone" when layers were actually pinned (warns otherwise).
- **dfine/rtdetrv4 invalid-anchor \`inf\`** → finite sentinel (this was missed; deim/deimv2/ec were already done). rtdetrv4 inherits D-FINE's decoder, so the one fix covers both. Verified: ec-s TRT-fp32 0.15 → 0.54.

Skipped with reason: the DAMO-YOLO pre-NMS cap (30000 ≫ \`max_det\`=300, so it cannot change the final detections — same logic as the verified picodet cap).

Re-review please:
@codex review
@cursor review
@coderabbitai review

PR #410, opened 2026-06-17
Auto-convert bare upstream RF-DETR checkpoints as COCO (nc=80)

EHxuban11:
## Problem

Loading a **bare upstream RF-DETR `state_dict`** (Roboflow's COCO-pretrained
weights — `class_embed.bias` shape `[91]`, no `names`/`nc`/dataset metadata)
auto-converts to **`nb_classes=90`** instead of `80`.

`_is_coco_rfdetr_checkpoint()` only recognized a COCO checkpoint via (a) exactly
80 names, or (b) a `"coco"` dataset metadata field. A bare metadata-less
checkpoint has neither, so it fell through to the conservative
"custom 90-class head" path and stayed `nc=90`.

## Impact

With `nb_classes=90`, RF-DETR's `_postprocess` **skips the COCO 91→80 class
remap** (the remap is gated on `nb_classes == 80`) and returns raw 91-class
labels. Any consumer that expects LibreYOLO's COCO-80 convention — e.g. COCO
mAP evaluation that maps 80→91 category IDs — then **double-maps every class**,
collapsing mAP to ~0.001 even though boxes and scores are correct.

This silently corrupts results: the model loads, runs, and emits confident,
well-localized detections — only the class IDs are wrong.

## Fix

Treat a **metadata-less 90-class RF-DETR as the canonical Roboflow COCO
checkpoint** — it is the only metadata-less 91-output RF-DETR in distribution —
so it normalizes to COCO-80, identical to the published LibreYOLO weights.
Genuine custom 90-class models are saved with `names`/metadata and are
unaffected (the change only triggers when `name_count is None`).

```python
# in _is_coco_rfdetr_checkpoint(), reached only when raw_nc == 90:
if name_count is None:   # bare state_dict, no class metadata at all
    return True
```

## Validation

COCO `val2017-mini500`, RF-DETR PyTorch, before → after:

| model | before | after | published nc=80 |
|---|---|---|---|
| rfdetr-n | 0.001 | **0.5135** | 0.5138 |
| rfdetr-s | 0.001 | **0.5512** | 0.5511 |
| rfdetr-m | 0.001 | **0.5732** | 0.5737 |
| rfdetr-l | 0.001 | **0.5855** | 0.5855 |

All four now match the published nc=80 weights within ±0.0005.

## Release Notes

* **Bug Fixes**
  * Improved auto-detection and normalization of RF-DETR COCO checkpoints to reliably recognize COCO-80 configurations.
  * Prevents unintended COCO fallback when checkpoints are custom/non-COCO 90-class variants.
  * Empty placeholder metadata values are no longer treated as dataset hints.

* **Documentation**
  * Added a “RF-DETR COCO normalization” guide clarifying detection and conversion rules.

* **Tests**
  * Expanded unit coverage for COCO-80 fallback vs 90-class preservation across different metadata and dataset-hint inputs.

### DIFF
--- a/docs/checkpoint_schema.md
+++ b/docs/checkpoint_schema.md
@@ -133,6 +133,19 @@ When metadata is missing or incomplete:
   checkpoints. Convert them with the appropriate `weights/convert_*.py` script
   before loading.
 
+### RF-DETR COCO normalization
+
+Upstream RF-DETR checkpoints expose a 91-output `class_embed` head
+(`raw_nc = 90`, COCO's 90 classes + background). Auto-conversion normalizes a
+*COCO* RF-DETR to LibreYOLO's COCO-80 convention (`nc = 80`, with the COCO
+remap applied at post-process). A checkpoint is treated as COCO when it carries
+80 names, a `coco` dataset hint, **or** no class/dataset metadata at all — a
+bare upstream state-dict is the canonical Roboflow COCO-pretrained checkpoint
+(the only metadata-less 91-output RF-DETR in distribution). A genuine custom
+90-class RF-DETR is preserved as `nc = 90`: it is identified by either a
+`names`/`class_names` list or a non-COCO dataset hint (e.g.
+`args.dataset_file`), so the bare-checkpoint COCO fallback does not fire for it.
+
 Schema helpers live in `libreyolo/utils/serialization.py`:
 
 ```python
--- a/docs/checkpoint_schema.md
+++ b/docs/checkpoint_schema.md
@@ -138,13 +138,20 @@ When metadata is missing or incomplete:
 Upstream RF-DETR checkpoints expose a 91-output `class_embed` head
 (`raw_nc = 90`, COCO's 90 classes + background). Auto-conversion normalizes a
 *COCO* RF-DETR to LibreYOLO's COCO-80 convention (`nc = 80`, with the COCO
-remap applied at post-process). A checkpoint is treated as COCO when it carries
-80 names, a `coco` dataset hint, **or** no class/dataset metadata at all — a
-bare upstream state-dict is the canonical Roboflow COCO-pretrained checkpoint
-(the only metadata-less 91-output RF-DETR in distribution). A genuine custom
-90-class RF-DETR is preserved as `nc = 90`: it is identified by either a
-`names`/`class_names` list or a non-COCO dataset hint (e.g.
-`args.dataset_file`), so the bare-checkpoint COCO fallback does not fire for it.
+remap applied at post-process). A checkpoint is treated as COCO when it:
+
+- carries exactly 80 names, **or**
+- declares an explicit class count of 80 (`nc` / `args.num_classes`), **or**
+- has a `coco` dataset hint, **or**
+- has **no** class or dataset metadata at all — a bare upstream state-dict is
+  the canonical Roboflow COCO-pretrained checkpoint (the only metadata-less
+  91-output RF-DETR in distribution).
+
+A genuine custom 90-class RF-DETR is preserved as `nc = 90`. It is identified
+by a `names`/`class_names` list, an explicit non-80 class count, or a non-COCO
+dataset hint (e.g. `args.dataset_file`), so the bare-checkpoint COCO fallback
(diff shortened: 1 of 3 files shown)


EHxuban11:
Thanks for the reviews — addressed in 9535aa7.

**Cursor & Codex (valid bug):** the bare-checkpoint fallback fired whenever `name_count is None`, even when a *non-COCO* dataset hint was present (e.g. a 90-class fine-tune saved with `args.dataset_file='custom90'` but no `names`). That would have mislabeled a genuine custom RF-DETR as COCO-80. Fixed: the fallback now fires only when there are **no names AND no dataset hint of any kind** — a dataset hint that isn't `coco` keeps the model at `nc=90`.

**CodeRabbit:** added two regression tests in `tests/unit/test_autoconvert.py`:
- bare state_dict (no args, no names), `raw_nc=90` → `nc=80`
- non-COCO dataset hint without names, `raw_nc=90` → `nc=90` (the case above)

…and documented the RF-DETR COCO-normalization rule under *Legacy And Foreign Weights* in `docs/checkpoint_schema.md`.

Verified all five branches (coco-hint→80, custom-names→90, bare→80, non-coco-hint→90, 80-names→80) resolve as expected.

@coderabbitai review
@cursor review
@codex review

EHxuban11:
Good catches again — both addressed in the latest push.

- **Non-string dataset hints** (CodeRabbit + Codex): any non-empty value under `dataset`/`dataset_file`/`dataset_name`/`data` now counts as a hint (only a *string* can positively identify COCO). A `data` config dict/Path no longer slips through.
- **Explicit class counts** (Codex): a top-level `nc` or `args.num_classes` now marks the class space as declared, so a 90-class fine-tune without names stays `nc=90`.

The bare-checkpoint COCO fallback now fires **only** when there are no names, no dataset hint of any type, and no explicit class count. Added regression tests for the dict-dataset-hint and `num_classes` cases.

Verified all branches: bare→80, coco-hint→80, 80-names→80, empty-hint→80, and custom-names / non-coco-string-hint / dict-hint / `num_classes` / top-level-`nc` → 90.

@coderabbitai review
@cursor review
@codex review

EHxuban11:
Thanks — round 3 addressed (commit pushed):

- **Empty placeholders** (CodeRabbit + Codex): a new `_has_metadata_value()` helper treats `""`, `{}`, `[]` as *absent*, so an otherwise-bare checkpoint with e.g. `data={}` still normalizes to COCO-80. Empty containers no longer block the fallback.
- **Explicit class count** (Codex): `nc`/`args.num_classes == 80` is now honored as COCO; any other declared count keeps the checkpoint at its own class space (so a 90-class fine-tune with `num_classes=90` stays `nc=90`).
- **Docs** (CodeRabbit): the COCO-normalization section now lists the explicit-class-count rule and the empty-placeholder handling.

Added tests for `data={}`/`[]` → 80 and `nc=80` → 80. Full matrix (13 cases) verified locally: bare/empty-container/`nc=80`/`num_classes=80`/`coco-hint`/`80-names` → 80; custom-names/non-coco-hint/dict-hint/`nc=90`/`num_classes=90` → 90.

@coderabbitai review
@cursor review
@codex review

EHxuban11:
Round 4:

- **Empty names placeholders** (Codex, valid): `names=[]` / `names={}` gives `name_count == 0` (not `None`), so the bare fallback didn't fire. Now treated as "no names" (`name_count in (None, 0)`) — symmetric with the empty-dataset-container handling. Test added.

- **"Explicit non-80 count should override 80 names"** (Codex): respectfully declining — this would *introduce* a regression. A legitimate COCO RF-DETR pairs LibreYOLO's 80 `names` with the model's **90 arch class count** (`num_classes=90`); letting the 90 override the 80 names would send a real COCO checkpoint to `nc=90`. The 80-names match is the authoritative COCO signal and correctly takes precedence (verified: `{names: <80>, nc: 90}` → 80). Stale 80-names on a re-headed custom model is a corrupt-checkpoint case I'd rather not optimize for at the cost of the common one.

Note: several inline comments this round are GitHub re-anchors of already-addressed findings (non-string hints, empty dataset containers) — those are handled in the current code; full 13-case matrix passes.

@coderabbitai review
@cursor review
@codex review

PR #437, opened 2026-06-26
Fix LibreFOMO release gaps: validator dispatch, exports, and docs

EHxuban11:
- Add `validator_class` hook to BaseModel so families can declare their own validator class; `val()` prefers it over the task-based dispatch.
- Set `LibreFOMO.validator_class = FOMOValidator` so `model.val()` uses FOMOValidator (with grid-level F1/P/R metrics) instead of the generic PointValidator.
- Make the models.fomo imports in fomo_validator.py lazy to break the validation ↔ models.fomo circular import that the above exposes.
- Export FOMOValidator from libreyolo.validation package.
- Fix nomenclature.md: remove the false "ships weights natively" claim; expand the point filename examples with m/l sizes; note no weights are auto-downloadable.
- Fix pre-existing test bug: threshold=0.25 causes all zero-logit cells to fire (softmax=0.5 > 0.25), producing 35 spurious FPs. Test now uses threshold=0.99 so only the explicitly boosted cell fires.

### DIFF
--- a/docs/nomenclature.md
+++ b/docs/nomenclature.md
@@ -160,7 +160,7 @@ only when it appears in that family's `SUPPORTED_TASKS`.
 
 Families that override `SUPPORTED_TASKS` also declare `TASK_INPUT_SIZES` so
 each task can use a different per-size input resolution (relevant for RF-DETR).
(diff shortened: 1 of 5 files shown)


EHxuban11:
@greptile-apps review

EHxuban11:
@greptile-apps review

Issue #484, opened 2026-07-02
Problems with background sample processing, mosaic/mixup performance, single/multi-GPU accuracy gap and lack of multi-class keypoint detection


kaiwen0901:
When using the code from this repository for our project, we encountered several issues listed below and would greatly appreciate your assistance with relevant explanations and optimizations:
1. When training YOLO9 on a dataset with background images accounting for more than 80% of all samples, its mAP@50 is roughly 12% lower than YOLOv8. After removing all background images from the dataset and retraining, its accuracy becomes comparable to YOLOv8. Similar situations have been found in multiple datasets. We want to understand how background images are handled in the current code, and wonder if the relevant processing logic can be optimized to maintain stable training accuracy on datasets with an overabundance of background samples.
2. When training YOLO9 on our custom dataset, the model accuracy stays extremely low with barely any improvement during mosaic and mixup augmentation. Accuracy increases drastically only after we turn off mosaic and mixup. Could you confirm whether this phenomenon is expected behavior?
The picture below is a screenshot of 100 epochs of training：
<img width="897" height="524" alt="Image" src="https://github.com/user-attachments/assets/650927b2-1d6e-4d73-a6db-5aa2e38f82fc" />

<img width="782" height="510" alt="Image" src="https://github.com/user-attachments/assets/ce0f8866-1cfa-43b1-a63c-1385fe9af26c" />

<img width="721" height="506" alt="Image" src="https://github.com/user-attachments/assets/d6b227ec-cf01-458f-baae-4b78978c0326" />

<img width="711" height="507" alt="Image" src="https://github.com/user-attachments/assets/fbe6a2a4-998c-4248-9227-280552e01167" />

<img width="694" height="522" alt="Image" src="https://github.com/user-attachments/assets/77766182-09b1-4d33-b343-21ca9c16d0c3" />

3. Under the exact same dataset and identical training configurations, we observed inconsistent final accuracy between single-GPU and multi-GPU training.

<img width="1375" height="557" alt="Image" src="https://github.com/user-attachments/assets/f5e15735-be6e-4ca9-a53d-21a6c0ef5902" />

4. All keypoint detection models implemented in the repo only support single-class detection. Would it be feasible to add code support for multi-class keypoint detection?

EHxuban11:
Hello @kaiwen0901 , thank you for the time you took to report this bugs, the experiments you did facilitated my debugging a lot. 

[...]

PR #488, opened 2026-07-02
Fix multi-GPU LR (incl. RT-DETR), mixup label-drop, background augmentation, and pose crash (#484)

EHxuban11:
Fixes the training-correctness bugs reported in #484, plus one regression caught by review during the work (RT-DETR DDP normalization). Status per reported problem is at the bottom.

## 1. Multi-GPU trained at ~N× effective LR (#484 point 3)

`scale_loss_for_ddp` multiplied every loss by `world_size`. That is only correct for a *pure-sum* loss, but **every** LibreYOLO loss is mean/ratio-normalized (yolo9 `cls_norm`, YOLOX `num_fg`, RTMDet/PicoDet `avg_factor`, YOLO-NAS `assigned_scores_sum`, the DETR families' `num_boxes`, classification `cross_entropy`, NAFNet per-pixel mean). DDP already **averages** gradients across ranks, so a normalized per-rank loss already composes to the single-GPU-equivalent gradient; the extra `× world_size` over-counted by ~N and inflated the effective learning rate. A 4-GPU run effectively trained at ~4× LR — matching the reporter's 4-GPU mAP 0.585 vs 1-GPU 0.814.

**Fix**
- `scale_loss_for_ddp` is now the identity (kept as the documented seam where a genuinely sum-reduced loss could opt back in; none currently exists). Its docstring now states the per-family normalizer contract explicitly.
- yolo9's `cls_norm` is globally reduced via a new `all_reduce_avg_scalar` helper (mirrors the DETR `num_boxes` reduction), through a single shared `YOLO9Loss._global_cls_norm` used by all four task losses, so multi-GPU matches single-GPU exactly (up to fp reassociation, ~4e-6). Single-GPU numerics are unchanged (`all_reduce_avg_scalar` == the old `max(sum, 1)` outside DDP).

**Proof** — new `tests/unit/test_ddp_loss_parity.py` compares the 2-rank gloo gradient to the single-GPU gradient on the *same global batch*. Ratio was **1.96 ≈ world_size** before, **1.00 / elementwise `allclose`** after.

### RT-DETR: review-caught inversion, fixed here

RT-DETR's `num_boxes` all-reduced the global count but — unlike D-FINE/DEIM/DEIMv2/RF-DETR — **never divided by world_size**. Under the old `× world_size` scaling the two errors cancelled and RT-DETR multi-GPU was accidentally correct; with the identity alone it would have trained at ~1/N effective LR. RT-DETR now uses the same `all_reduce_avg_scalar` reduction as everything else, restoring exact single-GPU equivalence. (Single-GPU numerics unchanged.)

The remaining local-normalizer families (YOLOX, RTMDet, PicoDet, YOLO-NAS) go from ~N× error to approximately-correct; making their normalizers globally reduced for bit-exact parity — and migrating the DETR families' inline reductions onto the shared helper — is a follow-up.

## 2. YOLO9 mixup dropped the second image's labels (#484 point 2)

`_mixup` vstacked the second image's labels onto the first image's already-padded `(max_labels, 5)` block and truncated back to `max_labels`, keeping only the first block — so every object from the second image (blended in at ~50% opacity) was dropped and trained as unlabeled background. Fix: strip padding from both, merge the real objects, re-pad (mirrors the shared YOLOX mixup). Regression test added.

## 3. Background images were never augmented (#484 point 1)

`YOLO9TrainTransform` early-returned on empty-label images, skipping HSV and flips entirely. On the reporter's dataset (>80% background) most of the training data was therefore never augmented: the model saw the exact same negatives every epoch while labeled images varied. The empty path now applies the same photometric/flip ops (no label sync needed). No golden parity fixture pins the yolo9 empty path and the labeled path is untouched — all 44 augment parity fixtures still pass. (The YOLOX-family empty path is intentionally left as-is: that family's contract is byte-parity with upstream YOLOX, which also skips.)

Honest scoping: we found **no single smoking-gun** for the reported −12% on background-heavy data. Three defects that plausibly compound with it are now fixed (mosaic geometric augmentation — #467, already on dev; mixup label-drop and background-augmentation skip — this PR). The reporter should retrain on dev and report back; if a gap remains, it gets its own investigation.

## 4. Pose `build_target` crash on crowded images

`reshape(len(kpts_px), -1)` used the uncapped instance count; when an image had more instances than `max_labels` the element count could not fill `len(kpts_px)` rows and numpy raised, crashing pose training. Fix: reshape the capped `n` instances. Regression test added. (Pre-existing; surfaced by Greptile on #483.)

## Status vs #484's four reported problems

| # | Reported problem | Status |
|---|---|---|
| 1 | Background-heavy dataset trains ~12% worse | Three compounding defects fixed (#467 on dev + points 2 & 3 above); no single root cause confirmed — retest requested |
| 2 | Mosaic/mixup stalls training | Mosaic fixed by #467 (dev); mixup label-drop fixed here |
| 3 | Single vs multi-GPU accuracy gap | Fixed here (yolo9 bit-exact, proven by gradient-parity test; DETR families corrected; RT-DETR normalizer fixed) |
| 4 | Multi-class keypoints | **Not in this PR** — investigated: all three pose families (RF-DETR/GroupPose, YOLO-NAS, EC) enforce nc=1 at head, loss, loader contract, and checkpoint schema, and both upstreams are person-only, so there is no low-risk plumbing path. Tracked as a feature request. |

## Behavior changes (release notes)

- Existing multi-GPU users were effectively training at ~N× LR. After this fix, N-GPU matches 1-GPU at the configured LR — the global batch is fixed and split across ranks, so no linear-LR-scaling rule applies. Users who previously compensated with a lowered LR should use the configured value. Gradient-clipping thresholds tuned on old multi-GPU runs now act on ~N×-smaller gradient norms; AMP's GradScaler re-adapts automatically within a few hundred steps when resuming old checkpoints.
- Background images now receive HSV/flip augmentation during YOLO9 training. Seeded runs consume the RNG stream differently than pre-fix versions (backgrounds now draw augmentation randomness), so per-sample augmentations are not reproducible *across* library versions — seeded runs within a version are unaffected.
- REVIEW.md's "DDP loss scales by world size" axiom updated to the new contract.

## Review pass

An 8-angle review (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions) ran over this PR before final push. It caught the RT-DETR inversion (fixed above) and drove: the shared `_global_cls_norm` method (was 4× copy-paste), a guard raising a clear error if the helper receives a CPU scalar under NCCL (was a latent hang), removal of a redundant `clone()`, and the REVIEW.md axiom update. Known accepted trade-offs, documented in code: the yolo9 loss now contains a collective (must run on all ranks — same property the DETR criteria already have), and the background/labeled augmentation blocks are kept separate with matching op order.

## Second review round (two independent agents on the final state)

A DDP-numerics specialist re-derived the gradient algebra for **every** training path and confirmed the PR's math (yolo9 exact to 3.9e-6 empirically, with negative controls: reinstating `× world_size` → ratio 2.000007; a local `cls_norm` → max diff 9.4e-2). It also confirmed DEIM/D-FINE never had the ×N bug (their trainers bypass the scaling seam), so this PR unifies a previously inconsistent framework.

An adversarial reviewer then found one real, measured defect in the new helper: it clamped the normalizer **after** dividing by world_size (the upstream-DETR order), so any global batch with positive mass below world_size — all-background batches above all — trained at ~mass/N gradient under DDP (measured ratio 0.5000 at 2 ranks). Fixed in the final commit by clamping the **global sum first**; parity is now exact for every batch including empty ones (a deliberate, documented deviation from the upstream-DETR clamp order). Two tests pin it: an all-background 2-rank gradient-parity test, and a 2-rank value test for `all_reduce_avg_scalar` covering the python-scalar + `device=` branch RT-DETR uses.

## Third review round (post dev-merge: merge audit + adversarial agent + Codex)

After merging dev (distillation #485, profiler #497) into the branch, three more independent reviews ran on the final state:

- **Merge audit: clean.** The merged `trainer.py` differs from dev's by comments only; all 11 PR files byte-identical to the branch parent. Bonus: dev's `_sync_distiller_grads` *averages* adapter grads — consistent with the identity scaling here (it was inconsistent with the old ×N), so the two PRs fix each other's interaction.
- **Adversarial agent:** found the HSV background test was flaky (~1/3 failure — `augment_hsv`'s draws are legitimately a visible no-op on an achromatic image). Replaced with a deterministic spy on the empty-label path. Also empirically verified RT-DETR fractional `num_boxes` DDP parity to 1.2e-7 including the denoising path, and that the all-background parity test cannot false-pass.
- **Codex (gpt-5.5, xhigh): one High finding, correct.** RF-DETR and EC pose still clamped their inline `num_boxes` *after* dividing by world_size; since both route through the shared trainer loop, removing the ×N compensation regressed their all-background corner to 1/world_size gradient. Both now use `all_reduce_avg_scalar` (clamp the global sum, then divide) — same fix as yolo9/RT-DETR. DEIM/D-FINE/DEIMv2 are intentionally untouched: they bypass `scale_loss_for_ddp` with their own epoch loops, so their pre-existing corner behavior is unchanged by this PR and stays a follow-up.

## Tests

- `test_ddp_loss_parity.py` — spawn gradient parity on a positive-rich batch AND an all-background batch, a 2-rank distributed value test for the reduction helper, plus single-process reduction-primitive tests.
- `test_yolo9_mixup_regression.py`, `test_yolo9_background_augment.py`, `test_pose_build_target.py`.
- Existing `test_ddp_distributed.py` DDP smoke tests still green (incl. RT-DETR's, which exercises the fixed `num_boxes` reduction under a real process group); stale ×world_size comments refreshed.
- Full unit suite green locally and on CI.

### DIFF
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -39,7 +39,7 @@ evidence-based, and scoped to the PR under review.
 - Per-rank loaders divide global batch.
 - Python multi-GPU training auto-spawns DDP.
 - Torchrun owns rank and device environment.
-- DDP loss scales by world size.
+- DDP losses are globally normalized; gradient averaging needs no world-size scaling.
 - Rank zero owns side effects.
 - Autobatch returns rank-divisible global batches.
 - Unit tests prove CPU-safe API behavior.
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -39,7 +39,7 @@ evidence-based, and scoped to the PR under review.
 - Per-rank loaders divide global batch.
 - Python multi-GPU training auto-spawns DDP.
 - Torchrun owns rank and device environment.
-- DDP losses are globally normalized; gradient averaging needs no world-size scaling.
+- DDP losses are mean-normalized (locally, or global-sum/world_size); gradient averaging needs no world-size scaling.
 - Rank zero owns side effects.
 - Autobatch returns rank-divisible global batches.
 - Unit tests prove CPU-safe API behavior.
--- a/libreyolo/data/augment/pose.py
+++ b/libreyolo/data/augment/pose.py
@@ -149,7 +149,7 @@ def build_target(
     target[:n, 1:5] = bboxes_px[:n]
     # Reshape the first ``n`` (capped) instances, not ``len(kpts_px)``: when an
     # image has more instances than ``max_labels`` the two differ and the old
-    # ``reshape(len(kpts_px), -1)`` raised a shape error (n*K*3 elements can't
-    # fill len(kpts_px) rows).
+    # ``reshape(len(kpts_px), -1)`` either raised a shape error or, when the
+    # element counts happened to divide, silently mis-broadcast coordinates.
     target[:n, 5:] = kpts_px[:n].reshape(n, -1)
     return target
--- a/libreyolo/data/augment/pose.py
+++ b/libreyolo/data/augment/pose.py
@@ -147,5 +147,9 @@ def build_target(
 
     target[:n, 0] = cls[:n]
     target[:n, 1:5] = bboxes_px[:n]
-    target[:n, 5:] = kpts_px[:n].reshape(len(kpts_px), -1)[:n]
+    # Reshape the first ``n`` (capped) instances, not ``len(kpts_px)``: when an
+    # image has more instances than ``max_labels`` the two differ and the old
+    # ``reshape(len(kpts_px), -1)`` raised a shape error (n*K*3 elements can't
+    # fill len(kpts_px) rows).
+    target[:n, 5:] = kpts_px[:n].reshape(n, -1)
     return target
--- a/libreyolo/data/augment/yolo9.py
+++ b/libreyolo/data/augment/yolo9.py
@@ -101,6 +101,19 @@ class YOLO9TrainTransform:
             padded_labels = np.zeros((self.max_labels, label_dim), dtype=np.float32)
             # Fill class with -1 to indicate padding (empty slots)
             padded_labels[:, 0] = -1
+            # Background images get the same photometric/flip augmentation as
+            # labeled images (no label sync needed — there are no labels). The
+            # old early return skipped HSV and flips entirely, so on
+            # background-heavy datasets most of the training data was never
+            # augmented and the model could memorize the exact negatives
+            # (issue #484). Same op order and RNG-draw pattern as the labeled
+            # path below.
+            if random.random() < self.hsv_prob:
+                augment_hsv(image)
+            if random.random() < self.flip_prob:
+                image = image[:, ::-1]
+            if random.random() < self.vertical_flip_prob:
+                image = image[::-1, :]
             image, _ = preproc(image, input_dim)
             if return_masks:
                 return (
--- a/libreyolo/data/augment/yolo9.py
+++ b/libreyolo/data/augment/yolo9.py
@@ -464,28 +464,44 @@ class YOLO9MosaicMixupDataset:
         return img, labels, (input_h, input_w), idx
 
     def _mixup(self, img, labels):
-        """Apply mixup augmentation."""
-        # Get another random image
-        idx2 = random.randint(0, len(self.dataset) - 1)
-        img2, labels2, _, _ = self._get_normal_item(idx2)
+        """Apply mixup augmentation.
 
-        # Mix images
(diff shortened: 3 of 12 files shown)


EHxuban11:
Superseded by #498 — identical commits, moved off the issue-numbered branch name. No review content lost; see #498.

PR #581, opened 2026-07-11
Fix work-package-2 review findings

EHxuban11:
# What does this PR do?

Fixes the findings from the deep review of `573-work-package-2`:

- **RTMDet-Ins boxes were never clamped** (correctness): all four clamp statements in `_postprocess_segment` used fancy indexing (`boxes[:, [0, 2]].clamp_(...)`), which clamps a copy. Boxes could be negative or extend past the input/original frame. Fixed with basic slicing plus a regression test that pins the exact clamped box.
- **Depth Anything 3 sky handling mixed batch statistics** (correctness): `_apply_mono_sky` computed the far-depth 0.99 quantile across the whole batch, but the depth validator batches independent images. Now computed per image, with a regression test asserting per-image far depths. Single-image behavior (and the parity goldens) are unchanged.
- **Export support matrix did not know the families merged in parallel**:
  - `rtmdet/segment`: explicit `blocked` entry matching the model guard (the fallback previously advertised "experimental" for a combination that always raises).
  - `swinir/restore/onnx`: explicit `experimental` entry with the measured reason. Verified empirically: the backend pads sub-canvas inputs before the transformer and window attention sees the padding; on a 24x32 input vs native inference, max drift is 195 grey levels with 26.6% of pixels off by more than one level. `docs/swinir.md` no longer calls this path plainly "supported".
  - `sam3`: added to the promptable-family blocks and to `inventory.py` `OPTIONAL_MODELS` (it was missing while SAM-1/2/MobileSAM were listed). Inventory snapshot and generated compat docs regenerated.
- **Removed the accidentally committed root `status.json`** training-monitor crash artifact and gitignored the path.
- **Published the missing SwinIR weights**: `LibreYOLO/LibreSwinIR{s,m,l}-restore` are now live on the HF org (5-file contract, Apache-2.0, upstream SHA-256s verified against PR #571's recorded values, converter `--verify` green). Cold-cache `LibreYOLO("LibreSwinIRm-restore.pt")` auto-download plus predict verified. The upload-skill family table and filename whitelist now include SwinIR, PPOCR, and the RTMDet `-seg` names.
- OCR runner comment now states that unknown detection-shaped kwargs are tolerated for generic callers (only `augment`/`tiling` are rejected), matching the actual behavior.

**Not addressed here (follow-ups):** `LibreRTMDet*-seg` weights are intentionally NOT published until a mask-mAP run against the official references exists (PR #572 records that none was possible); the clamp fix above changes segment postprocess outputs, so that run should happen after this merges.

# Provenance declaration (required)

- [x] All code in this PR was written by the author(s), OR every ported or adapted part is listed in the table below.
- [x] No part of this PR is derived from GPL, AGPL, LGPL, non-commercial, proprietary, or unknown-license code, including rewrites, paraphrases, or renamed adaptations of such code.
- [x] Notice files are updated if this PR ports third-party code. No new third-party code is ported by this PR, so notice updates are not required.

Ported or adapted code (leave the table empty if none):

| LibreYOLO file | Upstream repository | Commit | License |
|---|---|---|---|
|  |  |  |  |

## Code provenance

All changes are original repository-native fixes to code already reviewed into this branch. No third-party code was copied, ported, adapted, paraphrased, or derived.

# How was this tested?

- Full unit suite on the fix branch: **3275 passed, 58 skipped, 4 xfailed, 0 failed** (the +2 over the branch head are the new regression tests).
- New regression tests fail on the unfixed code (clamp no-op and batch-mixed sky) and pass on the fix.
- `tools/gen_compat_table.py --check` green via `test_generated_export_docs_are_current`; committed inventory pinned by `test_committed_inventory_matches_runtime_inventory`.
- SwinIR ONNX drift measured empirically on random weights to justify the experimental tier (padded 24x32: max 195, 26.6% pixels >1; exact 64x64 canvas: max 6).
- SwinIR weights: upstream SHA-256s match PR #571, converter `--verify` green for all three sizes, cold-cache auto-download + predict verified for `m`.
- Ruff check and format green on all changed Python files; `git diff --check` clean.


EHxuban11:
Follow-up from the PR description is now resolved: the RTMDet-Ins mask-mAP parity run was completed on full COCO val2017 (imgsz=640, conf=0.001, RTX 5070 Ti) with the clamp fix from this PR, and all five `LibreRTMDet{t,s,m,l,x}-seg` weight repos are published on the HF org.

Parity vs official mmdetection references (mask mAP50-95, ours / official): t 35.38/35.4, s 38.72/38.7, m 42.08/42.1, l 43.68/43.7, x 44.61/44.6. Box mAP matches equally (max delta 0.13 on s). Upstream checkpoint SHA-256s match the hashes embedded in mmdet's release filenames; converters reported clean 546/582/762/942/1122-tensor loads; cold-cache `LibreYOLO("LibreRTMDett-seg.pt")` auto-download + predict verified. Model cards carry the per-size measured-vs-official table.

With this, the `rtmdet/segment` support-matrix entry could be revisited (native inference is now parity-validated; the blocked entry correctly still applies to export only).

PR #600, opened 2026-07-13
Add OMDet-Turbo open-vocabulary detection

EHxuban11:
# What does this PR do?

Adds OMDet-Turbo as a public open-vocabulary detector on `dev`.

- Exposes `LibreOMDetTurbo` and the `omdet-turbo`, `omdet`, and `omdet-turbo-swin-tiny` aliases.
- Uses the maintained Transformers model, processor, and post-processing for public `predict()`, while retaining the existing parity-verified native model as groundwork for a possible future fixed-vocabulary export contract.
- Adds adapter-vs-raw-Transformers parity, gated native parity, vocabulary edge-case coverage, install-surface coverage, and nightly real-weight e2e coverage.
- Adds `timm` to the `openvocab` extra because OMDet-Turbo's configured Swin backbone is instantiated through `TimmBackbone`.
- Updates nomenclature, the open-vocabulary ADR/design/testing documentation, and family/root/weights notices.

ONNX export remains explicitly out of scope under ADR 0008.

## Code provenance

- [x] All code in this PR was written by the author(s), OR every ported or adapted part is listed in the table below.
- [x] No part of this PR is derived from GPL, AGPL, LGPL, non-commercial, proprietary, or unknown-license code, including rewrites, paraphrases, or renamed adaptations of such code.
- [x] Notice files are updated if this PR ports third-party code (`THIRD_PARTY_NOTICES.txt`, per-family `NOTICE`).

Ported or adapted code (leave the table empty if none):

| LibreYOLO file | Upstream repository | Commit | License |
|---|---|---|---|
| `libreyolo/models/omdet_turbo/nn.py` (already on `dev`; this PR adds its missing notice) | https://github.com/huggingface/transformers (`src/transformers/models/omdet_turbo/`) | `2ef31dec1676249d26044a8aa8abe33dbecf0d10` (v4.45.0 reference release; the exact checkout used during original authorship was not recorded) | Apache-2.0 |

The user-facing adapter and its initial tests/docs come from LibreYOLO commit `299cddf4`; the additional verification, packaging, documentation, and provenance work was written for this branch.

Architecture origin: https://github.com/om-ai-lab/OmDet at `956e2f36a13c32bd1e30b14a790233268c2305fb`, Apache-2.0.

Checkpoint source: `omlab/omdet-turbo-swin-tiny-hf` at revision `7fe93cecfb770c4d76cf71163956221249cab566`, Apache-2.0. The `LibreYOLO/LibreOMDetTurbot` mirror at `d569ff1da5668156f07b5859246be435804fed2d` is byte-identical for model/configuration/processor/tokenizer assets; `model.safetensors` SHA-256 is `439d1575d7e237ad565ed6969ea2a2dfcedf2086155e9ca3ac96cd6180a48cfd`.

# How was this tested?

- `uv run python weights/parity_omdet_turbo.py` — PASS; decoder coordinate/class maximum absolute differences were `6.080e-06` and `3.839e-05`.
- `uv run --no-sync pytest tests/unit/test_openvocab.py tests/unit/test_omdet_turbo_parity.py -q` — 41 passed, 1 gated parity test skipped.
- `$env:LIBREYOLO_OMDET_PARITY=1; uv run --no-sync pytest tests/unit/test_omdet_turbo_parity.py -q` — 1 passed.
- `uv run --no-sync pytest tests/e2e/test_openvocab_inference.py -q -m openvocab -k omdet` — 2 passed.
- `$env:LIBREYOLO_PR_GATE=1; uv run --no-sync pytest tests/unit -m "unit and not external_data and not network"` — 3350 passed, 52 skipped, 58 deselected, 4 xfailed.
- `uv build` — wheel and sdist built; wheel contents and `openvocab` dependency metadata verified.
- Targeted Ruff checks and `git diff --check` passed.
- `save=True` renders were visually inspected for the requested vocabularies. On the local CPU run, OMDet-Turbo averaged 0.715 s after warm-up versus 4.656 s for Grounding DINO.

Model-quality note: at the default `conf=0.3`, both the raw checkpoint and adapter produced one false-positive `giraffe` detection on the parkour image. At `conf=0.5`, both returned zero. The adapter was not changed to suppress a raw-model output.


EHxuban11:
## Review + fixes pushed (c8747416)

Verified against the published weights: cold autodownload from `LibreYOLO/LibreOMDetTurbot` works, and `set_classes(["dog", "bicycle", "pickup truck"])` on `tests/fixtures/dog.jpg` returns bicycle 0.947, dog 0.927, pickup truck 0.748 with correct boxes, including the multi-word label mapping. Default COCO-80 vocabulary, `conf=` and the `iou=` warning all behave as documented. The real `post_process_grounded_object_detection` signature matches the `text_labels=`/`nms_threshold=` kwargs the adapter passes (checked on transformers 5.13.1; the unit tests mock the processor, so nothing in the PR gate covers that contract).

### Fixed: the family was in the factory but in none of the surfaces its siblings are in

`grounding_dino` and `owlv2` are registered in three places `omdet_turbo` was missing from. Greptile independently flagged two of the three.

1. **`export/support.py`** — with no `_FAMILY_BLOCKS` entry, `get_support("omdet_turbo", "detect", "onnx")` fell through to **`experimental`** while both siblings return `blocked`, so the support matrix advertised an ONNX export that the tier's `export()` actually refuses. To be precise: this was a metadata inconsistency, not a runtime bug. I ran the export and it raises a clean `NotImplementedError`. Now blocked, consistent with the docs and ADR 0008.
2. **`models/inventory.py`** — unregistered, so the family never appeared in the compatibility inventory (Greptile's "Model Inventory Omits New Family").
3. **`ui/server.py`** — the open-vocab picker is a hardcoded list, so `omdet-turbo` was constructible in Python but not selectable in the UI (Greptile's "UI Discovery Omits New Alias").

Regenerated `reports/export_inventory.json`, `README.md` and `docs/export_support.md` from `tools/`; the family now reports `export_override: blocked` for every format. `test_committed_inventory_matches_runtime_inventory` would have caught a stale report; 64 unit tests in the affected files pass.

### Provenance gate

It was failing on a heading mismatch, not on missing content: the body had `# Provenance declaration (required)` while the CI gate greps for `## Code provenance`. Retitled the existing section; the declaration itself was already adequate.

### Greptile "Prediction IOU Is Ignored" — the premise is wrong, but there is a real question underneath

The claim is that differing `iou=` values "produce identical suppression behavior **without warning**". They do not: `LibreOpenVocabDetector.__call__` already warns, and I confirmed it at runtime:

```
UserWarning: LibreOMDetTurbo does not run LibreYOLO NMS; iou= is accepted for API compatibility but ignored.
```

So the option is explicitly disclosed, not silently swallowed, and that matches the documented tier contract.

The legitimate question it raises: unlike Grounding DINO and OWLv2, OMDet-Turbo's processor **does** take an `nms_threshold`, so this family alone *could* honour `iou=` instead of pinning 0.5. That is a deliberate contract change (it would need a per-family capability flag so the base stops warning for this one), so I have not done it unasked. Worth a follow-up decision.

### Still worth deciding: the native graph is now dead code

`OmDetTurboDetectionModel` (584 lines) is imported by nothing except the gated parity script now that `predict()` delegates to transformers. The new NOTICE is commendably candid that "the exact checkout used during authorship was not recorded". Upstream is Apache-2.0 so there is no licensing problem, but it leaves derived code with acknowledged-imprecise provenance that no user path executes. Keep it as export groundwork (as the docs say) or delete it with its parity script and re-add when export is actually built. My preference is to delete, but that is an architecture call.

EHxuban11:
## Greptile round 2 addressed (940e105f, plus the alias fold)

**"IOU remains silently ignored"** — fixed, and with exactly the suggested line (`nms_threshold=float(iou_thres)`). The threshold is now declared per family via `NMS_THRESHOLD` on the base: set means the family suppresses boxes itself and honours `iou=`; `None` keeps the warning for Grounding DINO and OWLv2, which suppress nothing. The family default is injected in `__call__` the same way `conf=` already is, because `predict()` defaults `iou=0.45` and forwarding that blindly would have silently retuned OMDet's NMS from 0.5 to 0.45 for every existing caller. Verified on real weights: the processor sees `0.5` when `iou=` is unset and `0.77` when `iou=0.77` is passed, and box counts move monotonically with the threshold (5 / 9 / 14 boxes at iou 0.01 / 0.3 / 0.9).

**"Inventory Emits Invalid Name"** — correct, and it was worse than reported. `libreyolo models` prints `<family>-<size>`, so it advertised `omdet_turbo-t` — but it also already advertised `grounding_dino-t` and `grounding_dino-b`, which the factory rejects for the same reason. That bug predates this branch; registering the family would have extended it rather than introduced it. Rather than paper over it with an underscore alias, the factory now folds underscores to hyphens, so every name the CLI prints loads as printed. Pinned with a test that walks the inventory rather than a hardcoded list, so a future family cannot regress it.

## Also in this round

The native `OmDetTurboDetectionModel` is deleted (~780 lines). With `predict()` delegating to transformers, nothing imported it except its own parity script: derived code, a NOTICE that admitted the upstream checkout used to write it was never recorded, and zero user paths executing it. Its parity script and test are gone with it, and the notices now describe only what LibreYOLO actually ships — an adapter over the Transformers implementation plus the mirrored Apache-2.0 checkpoint.

## Verification

Cold autodownload from an empty `weights/` against the published `LibreYOLO/LibreOMDetTurbot`: 4.2s, loaded via the CLI-advertised name `omdet_turbo-t`, and `set_classes(["dog", "bicycle", "pickup truck"])` returns bicycle 0.947, dog 0.927, pickup truck 0.748 with correct boxes. 67 unit tests pass in the affected files.

PR #602, opened 2026-07-13
Add EdgeTAM promptable segmentation

EHxuban11:
## What does this PR do?

- Add `LibreEdgeTAM` image promptable segmentation through the existing LibreSAM contract.
- Add factory aliases, public exports, inventory metadata, blocked export declarations, documentation, and notices.
- Add reproducible safe checkpoint conversion and guarded Hugging Face upload tooling for `LibreYOLO/LibreEdgeTAM`.
- Match the official EdgeTAM square resize, normalization, and prompt-coordinate transforms exactly.
- Keep video tracking and export explicitly out of scope for v1.
- Add hermetic logic tests, gated upstream parity, and real-weight E2E coverage.

## Code provenance

Runtime integration, factory wiring, offline-loading compatibility work, uploader safeguards, and tests are original code written for this PR.

EdgeTAM preprocessing behavior is adapted from `facebookresearch/EdgeTAM` at commit `7711e012a30a2402c4eaab637bdb00a521302c91`, Apache-2.0.

The checkpoint key-conversion mapping is adapted from `huggingface/transformers` at commit `bd37c453544e83eb875ed3608980a1660376007a`, Apache-2.0.

The generated mirror copies only hash-verified non-weight configuration and processor files from `yonigozlan/EdgeTAM-hf` revision `c266ce53b3fc00f0f495b583f6a116c4e57f53bb`, whose model card declares Apache-2.0. Learned tensors are independently converted from `facebook/EdgeTAM` revision `14d7ecc48c656b94e5184519f698cd5386c5a2bf` and checked exactly against that reference. The official raw checkpoint SHA-256 is `ed2d4850b8792c239689b043c47046ec239b6e808a3d9b6ae676c803fd8780df`.

All reused material and modifications are disclosed in the repository notice surfaces. No incompatible-license source was used.

## How was this tested?

- Hermetic PR gate: 3,277 passed, 129 skipped, 49 deselected.
- Nine-stage upstream parity: exact preprocessing, 984 exact tensors, maximum decoder-logit difference `5.149841e-05`, and point/box/grouped-prompt final mask IoU `1.0`.
- Real-weight EdgeTAM E2E: 3 passed.
- Fresh empty-cache offline loading passed without global config mutation or serialized workaround fields.
- Compatibility inventory, Ruff, targeted ty, lockfile/dependency checks, install smoke, upload dry run, and manual visual QA passed.
- CPU measurement: EdgeTAM image encoding was approximately 3.13x faster than LibreSAM2 tiny on the same machine.

## Before merge

- [ ] Publish `LibreYOLO/LibreEdgeTAM` and verify a clean-user autodownload; publication is pending a write-capable Hugging Face credential.


EHxuban11:
## Weights published + a provenance fix pushed

`LibreYOLO/LibreEdgeTAM` is live: https://huggingface.co/LibreYOLO/LibreEdgeTAM (commit `c8172893`). Cold autodownload from an empty `weights/` verified, and `tests/e2e/test_edgetam_smoke.py` passes 3/3 against the published repo.

Running the conversion end to end surfaced a bug that blocked it, fixed in 516ec335.

### Pinned upstream digests were CRLF-mangled

Every pinned `facebookresearch/EdgeTAM` digest was computed on a checkout with git's `core.autocrlf` translation active, so it hashed CRLF-translated bytes rather than the bytes upstream actually publishes:

| Artifact | was | now |
|---|---|---|
| `LICENSE` | `1eb85fc9…` | `c71d239d…` |
| `sam2/utils/transforms.py` | `c200d193…` | `ba3a64f4…` |
| `sam2/build_sam.py` | `ccb8fee6…` | `36406726…` |
| `sam2/sam2_image_predictor.py` | `edd7281e…` | `f13e5f9d…` |
| `sam2/configs/edgetam.yaml` | `251eca7a…` | `25c2fda8…` |
| `_UPSTREAM_TREE_SHA256` | `e93885c6…` | `a5ca617d…` |

CRLF-converting the real upstream bytes reproduces the old constants exactly, which is how it was diagnosed. Three consequences, all now gone:

1. `convert_edgetam_weights.py` without `--upstream-license` fetched the raw LF `LICENSE` and always failed its own SHA check, so the reproducible path could not be reproduced by anyone.
2. `tests/unit/test_edgetam_parity.py` would `pytest.fail` ("checkout is not pinned at 7711e012") on any LF checkout, i.e. on Linux CI.
3. The published card would have advertised a LICENSE digest that does not match the file on GitHub.

The checkpoint and reference-snapshot digests were unaffected (binary, fetched via the Hub, not git), which is why everything else validated.

### Conversion result (real weights)

```
checkpoint_sha256:       ed2d4850… (matches the pinned official edgetam.pt)
exact_reference_tensors: 984
model_sha256:            8858f8e4… (bit-identical to the pinned reference)
```

984/984 tensors exactly equal to the pinned reference, strict-loaded into `EdgeTamVideoModel`, payload LICENSE now byte-identical to upstream (11,357 bytes, LF).

The "Before merge" box is now satisfied. Separately: there is no SAM/promptable collection in the org, so this repo (like `LibreSAM2`, `LibreMobileSAM`, `LibrePicoSAM3`) currently belongs to none.

EHxuban11:
## Greptile review addressed (1f0295b1)

### P1 "Prompt Transform Mutates Caller Data" — not reachable, but hardened

Correct about `torch.as_tensor` sharing storage, but **not a live bug**: `predict()` routes every prompt through `normalize_points`/`normalize_boxes`, which go via `_to_list()` → `.tolist()` and hand `_encode` fresh Python lists. A caller's ndarray/tensor is already never aliased, so reusing an array across predictions does not drift today.

I confirmed this the hard way: I first wrote a regression test that called `_encode` directly with a tensor, and it failed *with* the clone applied, because a raw tensor never survives that path (the processor stub rejects it) — which is precisely the evidence that the reported scenario cannot occur through the public API.

Applied the `.clone()` anyway as defence in depth for direct `_encode` callers, and added a test asserting the property users actually care about (predict twice with one ndarray → identical masks, caller array untouched) rather than the mechanism.

### P1 "Import Forces Process Offline" — real, fixed

Agreed, and this one is genuine. The module is explicitly importable (`except ModuleNotFoundError: from weights.parity_edgetam import ...`), so setting `HF_HUB_OFFLINE`/`TRANSFORMERS_OFFLINE` at module scope would break unrelated `from_pretrained` calls in whatever imported it. Moved into `_force_offline()`, called at the top of `run_parity()`. `timm` and `transformers` are imported lazily inside the run, so the script's no-network guarantee is unchanged. Verified: importing the module now leaves both variables unset.

### Also added

Coverage for the segment-everything grid path, which the EdgeTAM `_encode` override had left untested (the base class slices prompt tensors that the override rebuilds).

### On the failing macOS check

`test_detr_cpu_export_matrix.py::test_detr_detect_raw_parity[onnx-LibreDFINE] — assert 0.94 > 0.95`. Unrelated to EdgeTAM: this PR touches nothing in the D-FINE or ONNX export path, and the same branch passed macOS at 190f5cf9 an hour earlier; the only change since was six hash constants. It reads as a borderline numeric threshold on macOS arm64. I have re-run the job to confirm.

PR #608, opened 2026-07-14
add tta for panoptic segmentation

imagra93:
# Add flip-TTA support for panoptic segmentation

## Code provenance

Original code written for this PR. No code was ported or adapted from any
third-party project. The dispatch/hook pattern (`_predict_augment_panoptic`,
the `_postprocess_*` split) follows the structure established by the
semantic segmentation flip-TTA PR (#601) already merged into this repo, not
an external source.

## Summary

Adds horizontal-flip test-time augmentation (`augment=True`) for panoptic
segmentation, following on from the semantic segmentation TTA support merged
in #601. Panoptic decode is query-based (Mask2Former/MaskFormer-style), so
the flip-merge strategy is architecture-specific and cannot reuse the
family-generic semantic hook — this PR adds a dedicated dispatch path plus a
mask-IoU deduplication step needed to make cross-view query fusion safe.

- `BaseModel._predict_augment` gains a `panoptic` dispatch branch and a
  `_predict_augment_panoptic` default hook (raises `NotImplementedError` for
  families that don't implement it, mirroring the semantic hook's pattern).
- Removed the two panoptic TTA rejection guards (`_predict_augment` and
  `val()`) that previously raised `ValueError("... does not support panoptic
  segmentation yet ...")`.
- `LibreEoMT._postprocess_panoptic` split into `_panoptic_queries` (decode
  raw output into filtered score/label/mask-probability queries) and
  `_fuse_panoptic_queries` (the existing winner-take-all per-pixel fusion) —
  behavior-preserving refactor that lets TTA reuse the decode step per view.
- `LibreEoMT._predict_augment_panoptic`: runs original + flipped views
  (atomically preprocess→forward→decode per view, since EoMT stashes
  per-call instance state that must not be interleaved across views), flips
  the flipped view's mask probabilities back into alignment, concatenates
  queries across views, deduplicates, then fuses.
- `LibreEoMT._dedup_panoptic_queries` (new): greedy, highest-score-first,
  mask-IoU-based merge of same-class queries above
  `PANOPTIC_TTA_DEDUP_IOU = 0.5` (mirrors the `PQ_IOU_THRESHOLD` convention).
- `PanopticValidator._run_validation_augmented`: per-image validation path
  for `augment=True`, reusing `model._predict_augment` (same pattern as
  `DetectionValidator`).

## Why the dedup step exists — an empirical finding, not a hypothetical

The naive concatenate-only prototype was measured against real EoMT-S
panoptic weights before the dedup step was added, and it *lost* detections
instead of improving them: 5 segments under `augment=True` vs 7 for
`augment=False` (and for each view fused alone), dropping categories 112 and
116 entirely.

Root cause: near-duplicate queries for the same real object, one from each
view, split the per-pixel winner-take-all roughly evenly. Each query's own
`won_area / own_area` ratio then fell below `PANOPTIC_OVERLAP_THRESHOLD`
(0.8), so *both* competing queries for that object were dropped by the
existing overlap-threshold logic — two views agreeing on an object made that
object disappear, not appear more confidently.

Adding `_dedup_panoptic_queries` (mask-IoU merge before fusion) fixed this:
the single-image sanity check below recovers all 7 segments, matching
`augment=False` exactly.

To confirm this wasn't a one-image fluke, the naive concatenate-only
prototype (dedup step monkeypatched to a no-op) was also run against the
full 5000-image COCO panoptic split. The single-image finding turned out to
understate the problem — without dedup, TTA doesn't just fail to help, it
substantially degrades panoptic quality across the whole validation set:

| metric | augment=False (baseline) | augment=True, no dedup | augment=True, with dedup (shipped) |
|---|---|---|---|
| PQ | 0.4674 | 0.3543 (−0.1131) | **0.4788 (+0.0114)** |
| SQ | 0.8153 | 0.7788 | 0.8147 |
| RQ | 0.5633 | 0.4478 | 0.5777 |
| PQ_things | 0.5015 | 0.3952 | 0.5152 |
| PQ_stuff | 0.4158 | 0.2925 | 0.4240 |

Every metric collapses without dedup, not just PQ_things (where cross-view
duplicate instances would be expected to bite) — PQ_stuff drops even more
in relative terms, since stuff categories are fused per-category and a
flood of near-duplicate low-confidence stuff queries competing in the same
winner-take-all pool degrades that fusion too. This is the concrete
evidence that the dedup step is load-bearing, not defensive-only code.

### Full COCO panoptic validation (`panoptic_val2017`, all 5000 images)

| metric | augment=False | augment=True | delta |
|---|---|---|---|
| PQ | 0.4674 | **0.4788** | **+0.0114** |
| SQ | 0.8153 | 0.8147 | −0.0006 |
| RQ | 0.5633 | 0.5777 | +0.0144 |
| PQ_things | 0.5015 | 0.5152 | +0.0136 |
| PQ_stuff | 0.4158 | 0.4240 | +0.0082 |
| time | 140.6s | 264.6s | ~1.9x |

RQ moves more than SQ, which is the expected signature of an ensembling
effect improving recall (fewer missed/duplicated instances) more than mask
precision — consistent with the dedup step's role.

### Single-image sanity check (`tests/fixtures/dog.jpg`, EoMT-S panoptic)

`augment=False` and `augment=True` both find **7/7 segments**, identical
categories `[1, 2, 16, 87, 100, 112, 116]`.

## Test plan

- [x] `pytest tests/unit/test_eomt.py tests/unit/test_panoptic_scaffold.py -m unit -q` — 86 passed, 3 skipped
- [x] `pytest tests/unit -m unit -q` (full suite) — 3278 passed, 197 skipped, 0 failures
- [x] New unit tests: `_dedup_panoptic_queries` merge/keep-separate cases (same class overlapping, different classes, non-overlapping same class), a regression test reproducing the exact empirical failure above, `_predict_augment_panoptic` query concatenation across views and single-empty-view handling, and the default `NotImplementedError` for families without an override
- [x] `PanopticValidator._run_validation_augmented` extraction and empty-result-handling tests
- [x] Real single-image sanity check against real EoMT-S panoptic weights
- [x] Full 5000-image COCO panoptic PQ validation, augment on vs off

### DIFF
--- a/libreyolo/models/base/model.py
+++ b/libreyolo/models/base/model.py
@@ -757,11 +757,6 @@ class BaseModel(ABC):
                 "Test-time augmentation does not support point-task models yet. "
                 "Use augment=False for point models."
             )
-        if getattr(self, "task", "detect") == "panoptic":
-            raise ValueError(
-                "Test-time augmentation does not support panoptic segmentation yet. "
-                "Use augment=False for panoptic models."
-            )
         if getattr(self, "task", "detect") == "depth":
             raise ValueError(
                 "Test-time augmentation does not support depth estimation yet. "
@@ -788,6 +783,16 @@ class BaseModel(ABC):
 
         if getattr(self, "task", "detect") == "semantic":
             return self._predict_augment_semantic(
+                img_pil,
+                image_path,
+                (orig_w, orig_h),
+                effective_imgsz,
+                color_format,
+                **kwargs,
+            )
+
+        if getattr(self, "task", "detect") == "panoptic":
+            return self._predict_augment_panoptic(
                 img_pil,
                 image_path,
                 (orig_w, orig_h),
@@ -897,6 +902,26 @@ class BaseModel(ABC):
             path=str(image_path) if image_path else None,
             names=self.names,
             semantic_mask=SemanticMask(mask.long(), (orig_h, orig_w)),
+        )
+
+    def _predict_augment_panoptic(
+        self,
+        img_pil,
+        image_path,
+        original_size: Tuple[int, int],
+        effective_imgsz,
+        color_format: str,
+        **kwargs,
+    ) -> Results:
+        """Flip-only TTA for panoptic segmentation. Override per family.
+
+        No family-generic implementation exists (unlike semantic's
+        ``_postprocess_semantic_logits`` hook): panoptic decode is
+        query-based (Mask2Former/MaskFormer-style), and the flip-merge
+        strategy for query outputs is architecture-specific.
+        """
+        raise NotImplementedError(
+            f"{type(self).__name__} does not implement panoptic flip-TTA."
         )
 
     def _merge_classify_tta(
@@ -1356,11 +1381,6 @@ class BaseModel(ABC):
             raise ValueError(
                 "Augmented validation does not support point-task models yet. "
                 "Use augment=False for point models."
-            )
-        if augment and self.task == "panoptic":
-            raise ValueError(
-                "Augmented validation does not support panoptic segmentation "
-                "yet. Use augment=False for panoptic models."
             )
         if augment and self.task == "depth":
             raise ValueError(
--- a/libreyolo/models/eomt/model.py
+++ b/libreyolo/models/eomt/model.py
@@ -692,14 +692,21 @@ class LibreEoMT(BaseModel):
         into it (mask probabilities and scores averaged) instead of
         competing against it separately in the fusion step.
 
-        ``view_ids`` (one int per query, matching ``scores``) restricts
-        merging to pairs from *different* views. Without it, two genuinely
+        ``view_ids`` (one int per query, matching ``scores``) restricts each
+        group to at most one query per view. Without it, two genuinely
         distinct same-class instances the model detected within a single
         view (e.g. two overlapping people in a crowd) could have high enough
         mask IoU to be wrongly folded into one segment — a risk this
         function only exists to dedup cross-view duplicates, not to
-        second-guess a single view's own instance separation. Pass ``None``
-        (the default) to merge on mask IoU alone, ignoring view origin.
+        second-guess a single view's own instance separation. The same cap
+        also stops one broad anchor query from absorbing *two* distinct,
+        correctly-separated queries from the opposite view: greedy
+        clustering otherwise has no limit on how many same-class queries a
+        single anchor can accumulate, so an imprecise mask that happens to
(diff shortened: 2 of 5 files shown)


imagra93:
Fixed in the latest commit: each query is now tagged with its source view (`view_ids`), and `_dedup_panoptic_queries` only merges pairs from *different* views. Added two regression tests (`test_dedup_panoptic_queries_never_merges_same_view_duplicates`, `test_predict_augment_panoptic_tags_queries_with_their_source_view`) — full suite still green (3280 passed, 0 failures).

Re-ran the full 5000-image COCO panoptic validation after the fix:

| metric | augment=False | augment=True, before fix | augment=True, after fix |
|---|---|---|---|
| PQ | 0.4674 | 0.4788 | **0.4772** |
| SQ | 0.8153 | 0.8147 | 0.8142 |
| RQ | 0.5633 | 0.5777 | 0.5759 |
| PQ_things | 0.5015 | 0.5152 | 0.5130 |
| PQ_stuff | 0.4158 | 0.4240 | 0.4231 |

Every metric ticks down slightly after the fix (PQ −0.0016, SQ −0.0005, RQ −0.0018, PQ_things −0.0022, PQ_stuff −0.0009), not just PQ — consistent with the bug inflating apparent quality across the board by occasionally merging distinct same-view instances into one segment. Against the `augment=False` baseline, TTA still improves PQ, RQ, PQ_things, and PQ_stuff (SQ dips slightly, −0.0011, same as before the fix — mask precision isn't what flip-TTA is expected to help), and that remaining gain is now attributable only to genuine cross-view averaging.

imagra93:
The view check added in the previous commit stops same-view merges, but
greedy clustering placed no limit on group size: a single anchor could
still absorb multiple same-class queries from the opposite view. A
broad/imprecise mask from one view overlapping two correctly-separated
instances from the other view (e.g. two people in a crowd) would merge
all three into one, silently losing an instance. Only manifests under
augment=True, since the non-TTA path never calls dedup.

Track which views are already represented in the current merge group
(group_views) and skip any candidate from a view already present -- so
a group holds at most one query per view, same cap that already applied
to the anchor's own view. Added
test_dedup_panoptic_queries_caps_one_match_per_view: one anchor with
identical masks to two same-class opposite-view queries now merges with
only the higher-scoring one.

Full-COCO PQ drops again slightly (0.4772 -> 0.4758) since some of the
remaining gain came from this over-merging too; +0.0085 PQ survives,
attributable only to genuine cross-view averaging.

Issue #631, opened 2026-07-22
Where can I find how to augment data for training in the doc?


juni3227:
This repository clearly can, and has code for image training-time-image augmentation, however, where is explanation for that? 
[data/augment/yolov9.py](https://github.com/LibreYOLO/libreyolo/blob/release/libreyolo/data/augment/yolo9.py)
I had re-read the doc multiple times, but .... how do I change settings for mosaic, affine transform, rotation, hsv, etc?

EHxuban11:
Hello @juni3227 , yeah, the docs are a mess in general.  I'll work on the augmentation section specifically. 

Thanks for reporting this. I'll keep you posted

EHxuban11:
Hello @juni3227 , the new releases (v1.4.0) docs now contains a augmentations section. 

https://www.libreyolo.com/docs/v1.4.0#augmentation  

Let me know if you have any other issues.

juni3227:
Thank you!

PR #635, opened 2026-07-22
Add a declarative augmentation spec; warn on ignored knobs and gated mixup

EHxuban11:
## What this does

Augmentation knobs on `TrainConfig` are shared across ~24 trainable families, but each family honors a different subset: DETR-style pipelines ignore `mosaic`/`mixup`/`hsv_prob`/the affine knobs, YOLO-NAS ignores `mosaic` but applies an always-on affine, classification families ignore all detection knobs, and in the YOLOX-style pipelines mixup and the affine knobs only fire on mosaic samples. None of this was machine-readable, so the CLI could only warn for RF-DETR (a hardcoded list), and setting an ignored knob anywhere else silently did nothing (context: the docs complaints in #631).

This PR makes the knob-support surface explicit and enforceable:

- **`libreyolo/data/augment/spec.py`** — a torch-free declarative table: for every trainable family (yolox, yolo7, yolo9/e2e/p2, yolonas, rtdetr/v2/v4, dfine, deim, deimv2, ec, rtmdet, picodet, rfdetr, fomo, resnet, convnext, mobilenetv4, efficientnetv2, dinov2, segformer, nafnet) and every `TrainConfig` augmentation knob, whether it is `used`, `gated_by_mosaic`, or `ignored`, with a note explaining why. Family-specific extras (`copy_paste`, `rot90`, `crop_resize_prob`, pose knobs) are listed separately. Multi-task families (EC, DINOv2) are marked conservatively: a knob is only `ignored` if it is ignored in *every* task, so a warning is never wrong.
- **Spec-driven CLI warnings** — `get_unsupported_train_params()` now derives ignored params from the spec for *all* families instead of returning a hardcoded list only for RF-DETR. `libreyolo train model=dfine-s ... mosaic=0.5` now prints `Warning: D-FINE ignores these parameters: mosaic`; same for YOLO-NAS, classification families, etc. RF-DETR's class-level non-augmentation ignores (`optimizer`, `momentum`, `nesterov`, `pretrained`) are unioned in unchanged.
- **Mosaic-gated mixup warning** — in the YOLOX-style pipelines mixup only applies to mosaic samples, so `mixup_prob>0` with `mosaic_prob=0` silently never fires. `BaseTrainer._setup_data` (and FOMO's data setup) now log a warning explaining exactly that. YOLO-NAS correctly does not warn (its mixup is independent of mosaic).
- **`tests/unit/test_augment_spec.py`** (29 tests) — pins the spec to reality so it cannot drift: sentinel-value plumbing through every family's `create_transforms()` and dataset wrapper (knobs claimed `used` must land on the transform/wrapper; knobs claimed `ignored` must not), wrapper-class checks for the mosaic-gating claims, CLI warning behavior end-to-end via `CliRunner`, spec↔`TrainConfig` field consistency, and registry coverage (every trainable family must have a spec entry, so adding a family without classifying its knobs fails the tests).

## No behavior changes to training

All additions are warning-only pure functions; nothing touches the augmentation math or the RNG stream, so trained results are byte-identical. Proof:

- Full unit suite on this branch: **3600 passed, 123 skipped, 22 deselected, 4 xfailed** (12m43s) — same skip set as `dev` on this machine (the golden `test_augment_parity` fixtures skip here because the local cv2 is 4.10 vs the 4.13 reference environment; they skip identically on unmodified `dev`, and none of the files they cover are touched by this PR).
- `ruff check` clean on all touched files.

## Follow-ups this enables (not in this PR)

Rendering the support matrix into the docs (the actual fix for #631's "which knob does what, where" complaint), and using the spec to generate truthful per-family `--help`/`cfg` output.

## Code provenance

All code in this PR is original, written specifically for this PR by reading LibreYOLO's own trainers, transforms, and dataset wrappers on `dev`. Nothing is ported, adapted, or derived from any third-party repository. In particular, no Ultralytics (AGPL) or other copyleft/unknown-license code was consulted or reproduced; the only overlap with other frameworks is generic knob *names* the codebase already uses (`mosaic`, `flipud`, `copy_paste`).

### DIFF
--- a/libreyolo/cli/commands/train.py
+++ b/libreyolo/cli/commands/train.py
@@ -522,16 +522,20 @@ def train_cmd(
             ),
         )
 
-    # RF-DETR: warn and ignore unsupported params
-    rfdetr_warnings = []
+    # Warn when explicitly-set params are ignored by the selected family
+    # (spec-driven; see libreyolo/data/augment/spec.py).
+    ignored_warnings = []
     unsupported_params = get_unsupported_train_params(family)
     if unsupported_params:
         for param_name in unsupported_params:
             if param_name in user_provided:
-                rfdetr_warnings.append(param_name)
-        if rfdetr_warnings:
+                ignored_warnings.append(param_name)
+        if ignored_warnings:
+            from libreyolo.data.augment.spec import display_name
+
             out.progress(
-                f"Warning: RF-DETR ignores these parameters: {', '.join(sorted(rfdetr_warnings))}"
+                f"Warning: {display_name(family)} ignores these parameters: "
+                f"{', '.join(sorted(ignored_warnings))}"
             )
 
     # Dry run: validate and show resolved config
--- a/libreyolo/cli/config.py
+++ b/libreyolo/cli/config.py
@@ -13,14 +13,35 @@ from libreyolo.tasks import task_to_suffix
 
 
 def get_unsupported_train_params(family: str | None) -> set[str]:
-    """Return the set of CLI parameters ignored by a model family's trainer."""
+    """Return the set of CLI parameters ignored by a model family's trainer.
+
+    Augmentation knobs come from the declarative spec
+    (``libreyolo.data.augment.spec``), which covers every trainable family.
+    Families may declare additional non-augmentation ignores via an
+    ``UNSUPPORTED_TRAIN_PARAMS`` class attribute (RF-DETR: optimizer/momentum/
+    nesterov/pretrained), which is unioned in.
+    """
+    from libreyolo.data.augment.spec import ignored_aug_params
+
+    from .aliases import TRAIN_ALIASES
+
+    internal_to_cli = {v: k for k, v in TRAIN_ALIASES.items()}
+    unsupported: set[str] = set()
+    for name in ignored_aug_params(family):
+        # Fields whose name collides with a CLI alias key (the classification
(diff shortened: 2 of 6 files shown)


EHxuban11:
Good catch (greptile P2): the coverage test used a second manual list, so a new family absent from both lists would pass. Fixed in 72740e2e - the required set is now derived from the registry (every family whose model class declares TRAIN_CONFIG, the repo's machine-readable trainable flag), plus an explicit legacy list for the four trainable families that predate the flag (dfine, ec, yolonas, segformer) and the lazily-registered rfdetr/dinov2. The legacy list is self-cleaning: an assertion fails as soon as one of those families gains TRAIN_CONFIG, so it can only shrink. Follow-up worth doing separately: declare TRAIN_CONFIG on those four model classes - its absence also means the CLI dry-run shows generic defaults for them instead of family defaults.

EHxuban11:
Addressed the P1 in e0ead31d: the coverage test now enumerates families through cli.config._iter_model_classes() - the exact discovery path the CLI uses (registry + lazy optional families) - instead of importing the registry directly and hardcoding rfdetr/dinov2 in the required set. A future lazy _ensure_* family has to be added to that iterator for the CLI to see it at all, and the moment it is, this test requires a spec entry for it. The only remaining leniency is scoped precisely: rfdetr/dinov2 spec entries are tolerated when their optional deps are absent (so the suite passes on minimal installs), but when the deps are installed they are discovered and enforced like everyone else.

PR #632, opened 2026-07-22
Mark LibreDFINE ONNX parity xfail on macOS host-class numeric drift

EHxuban11:
## Summary
- Marks the LibreDFINE ONNX parity e2e test as xfail on macOS, where host-class numeric drift makes the strict parity threshold flaky.
- Unblocks the intermittent red macOS lane seen during v1.4.0 release validation; no product code changes.

PR #636, opened 2026-07-23
Fix v1.4.0 release blockers: openvocab deps, v1.3 positional compat, LoRA export order, D-FINE macOS parity

EHxuban11:
## Summary

Fixes the four confirmed blockers from the v1.4.0 release audit of PR #634, plus the recurring macOS D-FINE ONNX parity failure.

### 1. `libreyolo[openvocab]` was broken as advertised
OV-DEIM always embeds prompts with the vendored CLIP BPE tokenizer, which imports `ftfy` and `regex` at predict time — neither was in the `[openvocab]` extra, so a clean openvocab-only install crashed on the first prediction (with a hint telling the user to install `[clip]`). The extra now ships both, the tokenizer hint names the right extra per model, and a dependency tripwire test pins this.

**Verified end-to-end:** built a fresh venv with `pip install -e .[openvocab]` only, downloaded LibreOVDEIM-s, and ran a real `predict()` — works.

### 2. `Results` v1.3 positional compatibility restored
v1.4 had inserted `panoptic` before `depth_map` and `matte`/`ocr`/`restore_scale` before `speed`/`track_id` in both `__init__` and `update()`, so v1.3-era positional calls silently bound `depth_map` to `panoptic` (crashing `summary()`). The new parameters now come after the complete v1.3 signature. All internal call sites pass keywords, so nothing else changes. Compat tests construct `Results` and call `update()` with the exact v1.3 positional argument lists and assert every binding.

### 3. `LibreEoMT` v1.3 positional compatibility restored
`num_queries` had been inserted before `device`, so `LibreEoMT(None, "l", 150, "cpu")` raised on `int("cpu")`. It now comes after the v1.3 parameters. Compat test added.

### 4. Export validates the format before merging LoRA adapters
`BaseModel.export` folded live PEFT adapters into dense weights *before* the format string was validated, so a typo'd format destroyed trainable adapter state and then raised. The exporter is now resolved (which validates the format) before the merge. Tests pin both the failure ordering (invalid format leaves adapters untouched) and the success ordering (validate, merge, export). RF-DETR's override delegates to `super().export()` and inherits the fix.

### 5. D-FINE ONNX macOS parity — aligned, not xfailed
`test_detr_detect_raw_parity[onnx-LibreDFINE]` failed repeatedly on macOS CI at 0.9467 vs the required 0.95 row match. Root cause: the graph selects queries with an in-graph top-k over this test's random-init (near-uniform) scores, so tiny host-class float drift reorders a handful of near-tied queries — whole rows then "mismatch" with no numeric parity loss. The test now applies the file's existing Hungarian query alignment (already used for LibreEC) before the row comparison. The numeric tolerance is unchanged, and a genuinely wrong box still fails after alignment. The strict xfails for the known DEIM/RT-DETRv2/v4 gaps are untouched and still xfail. This supersedes the blanket macOS xfail proposed in #632.

### Also: FOMO export parity test de-flaked
`test_exported_point_parity` built its random model from ambient RNG state, so its pass/fail depended on which tests ran (and drew randoms) before it — adding any test earlier in the suite could flip a near-tied point ordering. The weight draw is now seeded, same as the DETR parity matrix.

## Validation

- Full unit suite locally (Windows, Python 3.12): **3606 passed, 123 skipped, 4 xfailed, 0 failed**.
- Real OV-DEIM prediction in a clean openvocab-only venv (the exact scenario that crashed before).
- DETR CPU export parity matrix: 10 passed, 4 xfailed (strict), including `onnx-LibreDFINE` with alignment.
- New compat tests fail on the unfixed code and pass with the fix.

## Code provenance

All changes are original code written for this PR: dependency declarations in `pyproject.toml`, parameter reordering in existing LibreYOLO signatures, a control-flow reorder in `BaseModel.export`, and new tests. No third-party code was ported, adapted, or derived.

### DIFF
--- a/libreyolo/export/exporter.py
+++ b/libreyolo/export/exporter.py
@@ -306,6 +306,12 @@ class BaseExporter(ABC):
         Returns:
             Path to the exported model file.
         """
+        # Private plumbing (quantized_export): a callable run at the
+        # post-validation mutation point below, so caller-side mutations
+        # (reconstructing fp32 masters, enabling export mode) also wait for
+        # every request rejection.
+        pre_trace_hook = kwargs.pop("_pre_trace_hook", None)
+
         task = getattr(self.model, "task", "detect")
         if task == "depth":
             # Depth export uses the fixed-resolution dense contract: backends
@@ -383,17 +389,22 @@ class BaseExporter(ABC):
             int8,
         )
 
+        # ---- Post-validation mutation point ------------------------------
+        # Every request rejection above — format support, precision
+        # validation, option preflight, imgsz/path resolution — has fired;
+        # from here on the live model may be mutated.
+        #
         # A model fine-tuned with lora=True carries live PEFT adapter layers.
         # Fold them into dense weights so the traced graph is a plain model
-        # with no peft dependency. Deliberately the last step before the
-        # artifact is produced: the merge is destructive (adapters are folded
-        # and removed), so every request rejection above — format support,
-        # precision validation, option preflight, imgsz/path resolution —
-        # must fire while the adapters are still intact and trainable.
+        # with no peft dependency. The merge is destructive (adapters are
+        # folded and removed), so it must not run any earlier.
         from ..training.lora import merge_lora_adapters, module_has_lora
 
         if module_has_lora(self.model.model):
             merge_lora_adapters(self.model.model)
+
+        if pre_trace_hook is not None:
+            pre_trace_hook()
 
         Path(output_path).parent.mkdir(parents=True, exist_ok=True)
 
--- a/libreyolo/export/exporter.py
+++ b/libreyolo/export/exporter.py
@@ -383,6 +383,18 @@ class BaseExporter(ABC):
             int8,
         )
 
+        # A model fine-tuned with lora=True carries live PEFT adapter layers.
+        # Fold them into dense weights so the traced graph is a plain model
+        # with no peft dependency. Deliberately the last step before the
+        # artifact is produced: the merge is destructive (adapters are folded
+        # and removed), so every request rejection above — format support,
+        # precision validation, option preflight, imgsz/path resolution —
+        # must fire while the adapters are still intact and trainable.
+        from ..training.lora import merge_lora_adapters, module_has_lora
+
+        if module_has_lora(self.model.model):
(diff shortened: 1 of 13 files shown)


EHxuban11:
The Greptile P1 is correct: the quantized branch still merged adapters before quantized_export() ran its recipe/format gates. Fixed in c8c2bd6a — the merge now happens inside quantized_export after every rejection path (the fmt != 'onnx' check also moved above the reprepare_model mutation, which had the same shape), and BaseModel.export no longer merges on the quantized path. Regression test added: an int8-quantized model with live adapters exported to a rejected format (torchscript) raises with the adapters untouched.

EHxuban11:
Also correct — format validation alone wasn't enough; the exporter's option preflight (nms=True on a non-YOLO9 family, int8 support checks, imgsz/path resolution) still ran after the merge. Fixed in 61e0a3bf by moving the fold into BaseExporter.__call__ as the last step before the artifact is produced, so every rejection path in every format (and every caller, including quantized_export's ONNX path) leaves live adapters intact. Tests now pin the exact scenario from this comment (OnnxExporter(non-yolo9)(nms=True) raises with adapters untouched) plus the preflight-then-merge ordering on a successful export. Remaining known limitation: a failure during the actual trace/serialization still happens on merged weights — that's inherent to in-place merging and doesn't change model outputs, only detaches the adapters.

EHxuban11:
Both correct — fixed in b4792f60.

**PT remainder:** the fold now happens inside export_finalized_pt, on the deep copy it already works on. This is strictly better than reordering: the live model is never mutated on the pt path at all (its documented contract — 'the live model stays prepared and trainable' — now also holds for LoRA adapters, on success as well as rejection).

**Rejected ONNX export unpacking weights:** reprepare_model and _set_export_mode(True) now run via a private _pre_trace_hook that BaseExporter.__call__ invokes at its post-validation mutation point (the same place the LoRA fold moved to), preserving their original relative order. A request rejected by preflight or imgsz/path resolution leaves the finalized packed state untouched.

New tests pin all three invariants: pt remainder rejection leaves adapters untouched, pt export folds on the copy only (merged model is not the live model), and a rejected quantized ONNX export keeps manifest state 'finalized'. Full unit suite locally: 3611 passed, 0 failed.

PR #649, opened 2026-07-27
Adding rectangular input support for convolution based detectors and etc.

juni3227:
Adding rectangular input resolution for libreyolo on convolution based models.

(YOLO9, 7, nas, rtmdet, picodet)

basically, it now accepts input as tuple[int, int] for imgsz and will construct models based on that. If models are not supported, error should be printed.

# Why this is needed
There is SAHI, and other tiling inference methods for large size images. However, there are cases where using unconventional resolution is still better then tiling method out of deployment reasons, such as pushing inference speed to the limit is a required. Also, there are cases when tiling inference would deter performance due to its inherent limit, or its application would rather make it slower.
Say, if you are using highspeed panorama camera (like, 1080x10800), thinner but wide aspect ratio would make tiling more computationally heavier if you use SAHI, and you would just want to train a new network with an odd ratio.

I made this commit because looking at the code base, only thing blocking us from using it is very odd choice of type of `imgsz` and nothing more.

Metric | Option A: SAHI (1024×1024) | Option B: Native Wide (1024×8192)
-- | -- | --
Number of Model Passes | 20 | 1
Processed Pixels per Pass | 1,048,576 px | 8,388,608 px
Total Pixels Evaluated | 20.97 Million | 8.38 Million
Overlap Redundancy | ~150% redundant area re-calculated due to tile overlaps | 0% redundant area

Validated on training, validation, inferencing. 

Many thanks to you.

## Code provenance
Created using Opencode, reviewed manually.

Rectangular image-size support is propagated through supported convolutional detection models.
- Parses scalar and `HxW` image sizes across train, validation, and prediction CLI commands.
- Updates preprocessing, postprocessing, validation, autobatch, and checkpoint metadata to preserve separate height and width dimensions.
- Rejects rectangular inputs for unsupported families and tasks, including YOLO-NAS.
- Adds unit coverage for rectangular parsing, validation, coordinate scaling, preprocessing, and capability gates.
</details>

The PR appears safe to merge.

No blocking failure remains; the prior model-native prediction-size issue is fixed by preserving an omitted size as None, and unsupported rectangular YOLO-NAS inputs are now rejected before cropping can occur.
</details>

| Filename | Overview |
|----------|----------|
| libreyolo/cli/commands/predict.py | Preserves an omitted image size as `None`, allowing inference to resolve the model-native input dimensions. |
| libreyolo/models/yolonas/utils.py | Rejects unsupported rectangular YOLO-NAS canvases and consistently caps resize dimensions for smaller square canvases. |
| libreyolo/postprocess/yolonas.py | Mirrors YOLO-NAS preprocessing resize and padding calculations when mapping detections and keypoints back to source coordinates. |
| libreyolo/training/trainer.py | Gates rectangular training by model family, task, and stride while preserving rectangular checkpoint metadata. |
| libreyolo/validation/detection_validator.py | Propagates separate validation height and width values through dataloading and ground-truth coordinate restoration. |
| tests/unit/test_rectangular_imgsz.py | Adds focused coverage for rectangular image-size parsing, configuration, preprocessing, postprocessing, and support restrictions. |

</details>

<sub>Reviews (3): Last reviewed commit: ["Add unit tests for rectangular imgsz: pa..."](https://github.com/libreyolo/libreyolo/commit/4fa78b7583b3458e0099685c0766da165e6a63b6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47472174)</sub>


juni3227:
There is another unresolved type change that is causing loading data from YOLO format dataset when you use rectangular size.

        # Scale to target image size
        r = min(self.img_size[0] / height, self.img_size[1] / width)
        if len(res) > 0:
            res[:, :4] *= r

https://github.com/LibreYOLO/libreyolo/blob/4515f75a1cd6b8977ee8aa6ed9b6d8e733dc63af/libreyolo/data/dataset.py#L486

I am opening other pull request.

PR #712, opened 2026-08-03
Lift kernel registry to libreyolo/kernels; opt-in hub MSDA kernel for DETR families

EHxuban11:
Opened by an agent. Verified: full CPU unit runs listed below. Not verified: the CUDA hub-kernel path (this machine has CPU-only torch).

## What changed

- Kernel registry lifted from `libreyolo/quant/kernels/` to library-wide `libreyolo/kernels/`, organized by purpose:
  - `kernels/quant/simulate/`: fake-quant Triton kernels (any device, STE backward; serve QAT and simulated PTQ/val).
  - `kernels/quant/execute/`: finalized-only real-precision paths (`scaled_mm_fp8`, fused fp8 cast/epilogue split out of `fp8.py` into `fp8_fusion.py`, `unpack_*`).
  - `kernels/attention/`: new `ms_deform_attn` op slot.
- New accelerated multi-scale deformable attention: with `pip install libreyolo[hub-kernels]`, the compiled CUDA MSDA kernel from Hub repo `kernels-community/deformable-detr` (Apache-2.0) serves RF-DETR, LibreDeformableDETR, and LibreDINO-DETR, forward and backward. Installing the extra is the opt-in (no package, no change, no network access); `LIBREYOLO_HUB_KERNELS=0` disables it. Eligible inputs are eager CUDA fp32; tracing/ONNX/torch.export always use the portable grid_sample path, so exports are untouched; load or runtime failures disable the provider for the process and fall back with one warning.
- `LIBREYOLO_KERNELS` supersedes `LIBREYOLO_QUANT_KERNELS` (legacy alias still honored). `libreyolo.quant.kernels` stays a working alias via a lazy shim.
- Docs: new `docs/kernels.md` (incl. per-platform behavior: CPU/MPS fall back to reference ops, ROCm can engage Triton); `docs/quantization.md` pointers updated. CHANGELOG and THIRD_PARTY_NOTICES updated.
- Why shared-code change: the slot is consumed by three families; the registry mechanism already existed and only moved.

## Reviewer notes

- The per-family MSDA ports stay byte-equivalent fallbacks; call sites only gained a guarded slot lookup that resolves to None by default.
- Import cycle `libreyolo.kernels` <-> `libreyolo.quant` is handled by a PEP 562 shim; `tests/unit/kernels/test_registry.py` covers the kernels-first import order in a subprocess.
[...]

PR #713, opened 2026-08-03
Kernel campaign: MSDA wire-ups for 8 more families, SDPA across 15

EHxuban11:
Executes the kernel campaign: Phase 1 (MSDA wire-ups) and Phase 2 (SDPA swaps). Phase 3 untouched.

Targets `dev`. #712 has since merged, and `dev` is merged in here (bd590b49) — the `ms_deform_attn.py` conflict was #712's new `is_exporting` guard landing on the same block this PR refactored into `ms_deform_attn_available()`; both are kept.

One family per commit. Happy to split into per-family PRs if preferred.

## What

**Phase 1 — 8 more families on the `ms_deform_attn` slot.** `lwdetr`, `grounding_dino`, `rtdetr`, `rtdetrv2`, `dfine` (+`rtdetrv4`), `deim` (+`deimv2`), `ec`, `openvocab/ovdeim`. Every one keeps its own grid_sample port as the default and the export path; the slot only takes over when a compiled implementation is eligible.

**Phase 2 — SDPA across 15 families.** Default-on where the parity bar is a tolerance, opt-in where it is byte-exact.

## The main finding: most Phase-2 families had to be opt-in

The plan expected "default-on likely fine" for most of Phase 2. Reading the actual bars, **8 of 14 attention modules are pinned byte-exact**, several against a reference whose *own* fused path is explicitly switched off to get there (`test_swin_parity.py:30` loops over the timm reference setting `fused_attn = False`; `parity_owlv2.py:32` says the same about transformers). Fused kernels accumulate in a different order, so per the plan's decision rule those keep manual math by default.

| Family | Bar | Where | Route |
| --- | --- | --- | --- |
| segformer | none found (searched `tests/`, `weights/`) | — | default-on |
| depth_anything (+moge2) | PSNR > 40 vs export | `tests/unit/test_depth_anything.py:216` | default-on |
| bert | `< 1e-4` | `weights/parity_bert.py:40` | default-on |
| grounding_dino | `< 1e-3` | `weights/parity_grounding_dino.py:108` | default-on |
| swinir | `rtol/atol 1e-3` vs export | `tests/unit/test_swinir.py:212-217` | default-on |
| ppocr | `< 1e-3` | `weights/parity_ppocr.py:104` | default-on |
| swin | **`== 0.0`** | `tests/unit/test_swin_parity.py:44` | opt-in |
| dinodetr swin | **`== 0.0`** | `tests/unit/test_dinodetr_parity.py:265` | opt-in |
| birefnet (+feynobg) | **`== 0.0`** | `weights/parity_birefnet.py:93` | opt-in |
| owlv2 | **`objectness_logits: 0.0`** | `weights/parity_owlv2.py:65` | opt-in |
| lwdetr (ViT + decoder MHA) | **`== 0.0`** | `weights/parity_lwdetr.py:150` | opt-in |
| siglip2 towers | **`_EXACT_TOL = 0.0`** | `tests/unit/test_siglip2_parity.py:46` | opt-in |
| zipdepth | **`== 0.0`** | `tests/unit/test_zipdepth_parity.py:84` | opt-in |
| mobilesam | **`== 0`** | `tests/unit/test_mobilesam_parity.py:73` | opt-in |

No tolerance was relaxed, no parity test edited or xfailed. Opt-in modules carry `fused_attn = False` (timm's convention, mirroring `vit/nn.py`); the switch is `libreyolo.kernels.attention.set_fused_attention(model)`. No new env vars.

siglip2's MAP pooling head is left entirely alone, as instructed.

## Measured diffs

**Default-on families, fixed-seed CPU forward, this branch vs #712 head** (random weights, whole-model output):

| family | max_abs_diff | output range | relative |
| --- | --- | --- | --- |
| segformer b0 | 5.96e-08 | 0.2485 | 2.4e-07 |
| swinir x2 | 7.15e-07 | 4.0125 | 1.8e-07 |
| ppocr rec t | 1.86e-09 | 0.0021 | 8.9e-07 |
| depth_anything vits | 2.61e-08 | 0.0349 | 7.5e-07 |
| bert encoder | 8.35e-07 | 6.7538 | 1.2e-07 |

fp32 rounding only. grounding_dino is not in this table (needs a full config to build); its text backbone is the bert row and both its changes are covered by the unit tests.

**Whole-model outputs that must not move at all** — same A/B, CUDA: `dfine_s`, `rtdetr_r18`, `deim_s`, `swin_t_cls` are **byte-identical** to #712 head. That is the Phase-1 wire-ups being inert without a compiled kernel, and swin's SDPA being off by default.

**GPU verification, RTX 5070 Ti.** CPU tests only exercise SDPA's math backend; on CUDA torch picks flash/mem-efficient/cuDNN, which is where mask handling actually differs. All 28 module/mask combinations agree in fp32 (worst 2.97e-06 relative, on a masked window-attention case) and under fp16 autocast (worst 1.3e-02 relative, fp16 noise). Speedups, fp16 autocast:

- Swin window attention, 512 windows x 49 tokens x 384: 1.278 ms -> 0.721 ms (**1.77x**)
- OWLv2 vision attention, 3600 tokens x 1024: 6.483 ms -> 1.735 ms (**3.74x**)

## Ambiguities, and what was done

- **BERT's boolean mask is a bias, not a gate.** `bert/nn.py` passes Grounding DINO a 4-D *bool* mask that the eager path ADDS (True -> +1, False -> +0), deliberately replicating a transformers quirk. SDPA reads a bool mask as "allowed" and would turn False into -inf — a different model, not a rounding difference. The mask is cast to the query dtype before SDPA sees it, with a regression test that fails if it is ever read as a gate.
- **Window masks.** Swin/SwinIR/BiRefNet/DINO-DETR add a relative position bias *and* a shifted-window mask; SDPA takes one additive mask, so they are summed and the window mask is tiled to match the (batch, num_windows) flattened batch. Every masked variant is covered by an equivalence test on CPU and GPU.
- **v2 ragged `num_points_list`** does not reshape onto the slot's `(n_levels, n_points)` layout: guarded with an explicit uniformity check, falls through. Tested.
- **`method='discrete'`** is integer-index sampling, a different equation: never reaches the slot. Tested.
- **dfine's forced-manual grid_sample** (`_FORCE_MANUAL_GRID_SAMPLE` / `_EXPORT`, the ONNX and Core AI escape hatch) outranks the slot. Tested. `_grid_sample_bilinear_manual` untouched.
- **dfine/deim/ec hand their core a per-level value list**, not the single tensor the slot wants, so the accelerated path pays one concat to rebuild it. Only on the accelerated path; unmeasurable here (see below).
- **ec's pose MSDA variant** left untouched, per the plan.
- 4 small helpers were added to `libreyolo/kernels/attention/ms_deform_attn.py` (availability pre-check, spatial-shape normalisation, the v2 layout adapter) rather than duplicating the same subtle uniformity check across 7 families.

## Tests

- `pytest tests/unit -m "not network and not external_data"` — **5033 passed, 1 failed** on the merge commit; the failure is the pidnet/ncnn flake below. (Earlier run on the pre-merge branch: 4989 passed, 2 failed. `dev` at the time, same command: 4863 passed, 0 failed.)

  The 2 failures are pre-existing flakes, not regressions. Evidence for each, rather than an assertion:

  - `test_pidnet.py::...test_exported_semantic_parity[ncnn]` — **fails 1 run in 6 on #712 head, running completely alone.** The mechanism is in the test and the exporter already: the test's own comment (`test_pidnet.py:219-222`) says untrained logits are near-tied so a rounding difference flips many argmax pixels, and the failing runs carry `ncnn.py:123`'s warning that the Windows PNNX wheel choked on an unescaped `C:\...` path in its generated loader and the artifacts were accepted anyway. Neither depends on this PR.
  - `test_cuda_graph_training.py::TestCudaParityYolo9::test_loss_trajectory_identical` — order-dependent: running `test_cuda_graph_training.py + test_pidnet.py` together, **#712 head fails it and this branch passes it**. It also passed on #712 head in the full suite and failed here, i.e. it moves in both directions independently of the change.

  Neither `yolo9` nor `pidnet` imports any module this PR touches.

- `tests/unit/kernels/` — 150 passed, 18 skipped (Triton kernels; no Triton on Windows). Two new files: per-family MSDA adapter tests (layout the slot receives, portable-path fallback, and a numerical-closure test running the adapted arguments through a reference MSDA) and per-module SDPA tests (both paths agree under every mask shape, export takes the primitive-op path even when opted in, `state_dict()` keys unchanged).
- `ruff` on every changed file: no new findings (5 pre-existing ones left alone).

Not verified:
- **The compiled Hub kernel itself.** `kernels-community/deformable-detr` builds are Linux-only and this box is Windows, so `ms_deform_attn_available()` is False here and the accelerated path never executed. What is verified is the contract: the shapes each family hands the slot, and that those arguments reproduce the family's own output when run through a reference MSDA. Someone should run one DETR family on Linux with `libreyolo[hub-kernels]` before this is trusted in anger.
- Weight-backed parity harnesses (`weights/parity_*.py`, `network`/`external_data` tests) were not run — they need checkpoints and network. They are the reason the byte-exact families are opt-in, so their defaults are unchanged by construction, but they have not been executed.

## Code provenance

All changes are original code written for this PR, plus rewiring of existing in-tree LibreYOLO code. No code was copied, ported, adapted, or derived from any third-party source. No new dependencies; the SDPA swaps use stock `torch.nn.functional.scaled_dot_product_attention`.

Existing upstream attribution headers in the touched files (Deformable-DETR, RT-DETR, D-FINE, DEIM, LW-DETR, DINOv2, MobileSAM and the rest, all Apache-2.0/MIT/BSD) are unchanged and still accurate: the ported math is preserved as the default path in every file, with the accelerated branch added alongside it.

The PR extends the shared deformable-attention slot to additional DETR-lineage families and introduces default-on or opt-in SDPA according to each family’s parity requirements.
- Adds shared MSDA layout and availability adapters with portable fallbacks.
- Adds graph-capture-aware SDPA policy and a model-wide fused-attention switch.
- Rewires attention implementations across transformer families and adds adapter, mask, parity, capture, and state-dict tests.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| libreyolo/kernels/attention/ms_deform_attn.py | Adds reusable availability, shape-normalization, and uniform-v2-layout adapters while preserving export and unsupported-layout fallbacks. |
| libreyolo/kernels/attention/sdpa.py | Centralizes graph-capture policy and provides fused-attention discovery and switching for bare modules and public model wrappers. |
| libreyolo/models/bert/nn.py | Routes eager self-attention through SDPA while preserving BERT’s unusual additive boolean-mask semantics and primitive export path. |
| libreyolo/models/grounding_dino/nn.py | Adds SDPA and compiled deformable-attention routing with portable fallbacks. |
| libreyolo/models/vit/nn.py | Makes the existing fused-attention flag functional without changing the default eager or ONNX behavior. |
| libreyolo/models/deit/nn.py | Makes the existing fused-attention flag functional while preserving the default SDPA and opset-17 export path. |
| tests/unit/kernels/test_ms_deform_attn_families.py | Exercises family-specific MSDA layouts, unsupported cases, fallback behavior, and numerical closure. |
| tests/unit/kernels/test_sdpa_attention.py | Covers default and opt-in attention routes, mask semantics, capture gates, wrapper switching, and unchanged state dictionaries. |

</details>

```mermaid
flowchart TD
    Call[Model attention call] --> Kind{Attention type}
    Kind -->|MSDA| Eligible{Compiled slot eligible?}
    Eligible -->|No or export| Portable[Family portable grid_sample path]
    Eligible -->|Yes| Adapter[Normalize family layout]
    Adapter --> Kernel[Compiled MSDA kernel]
    Kernel -->|Load or runtime failure| Portable
    Kind -->|Scaled dot-product| Capture{ONNX or jit.trace capture?}
    Capture -->|Yes| Manual[Primitive MatMul Softmax path]
    Capture -->|No| Policy{Family parity policy}
    Policy -->|Default-on| SDPA[PyTorch fused SDPA]
    Policy -->|Byte-exact default| Flag{fused_attn enabled?}
    Flag -->|No| Manual
    Flag -->|Yes| SDPA
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Merge remote-tracking branch &#39;upstream/d..."](https://github.com/libreyolo/libreyolo/commit/bd590b49b4f357596465f7cf3df397052cafb250) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49871691)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,15 +30,23 @@ before 1.4.0 are documented in the
 
 - Fused scaled-dot-product attention across the transformer families, using
   stock torch (no optional dependency). SegFormer, Depth Anything (and
-  MoGe-2), BERT, Grounding DINO, SwinIR and PP-OCR use it by default; ONNX
-  export still traces the primitive-op equation, so exported graphs are
-  unchanged. Families pinned to a byte-exact parity bar (Swin,
+  MoGe-2), BERT, Grounding DINO, SwinIR and PP-OCR use it by default; graph
+  capture (ONNX, plus the `torch.jit.trace`-based TorchScript, CoreML and
+  NCNN exporters) still records the primitive-op equation, so exported graphs
+  are unchanged. Families pinned to a byte-exact parity bar (Swin,
   LibreDINO-DETR's Swin backbone, BiRefNet, FeyNoBG, OWLv2, LW-DETR,
   SigLIP 2, ZipDepth, MobileSAM) keep manual attention by default and opt in
   with `libreyolo.kernels.attention.set_fused_attention(model)`, which trades
   byte-exact agreement with upstream for the fused kernels. Measured on an
-  RTX 5070 Ti under fp16 autocast: 1.77x on Swin window attention, 3.74x on
+  RTX 5070 Ti under fp16 autocast: ~2x on Swin window attention, ~3.8x on
   OWLv2 vision attention. See `docs/kernels.md`.
+
+### Fixed
+
+- `LibreViT` and `LibreDeiT` now honor their `fused_attn` attribute. It was a
+  timm-compatibility vestige that nothing read, so
+  `libreyolo.kernels.attention.set_fused_attention` reported switching it
+  while fused attention kept running. Default behavior is unchanged for both.
 
 - `val_loss=True` extended from the `g0` flagships to **every trainable
   family** (`g0`, `g1` and `g2`), across four tasks rather than detection
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,13 +17,28 @@ before 1.4.0 are documented in the
   and `libreyolo.quant.kernels` remains a working alias. See
   `docs/kernels.md`.
 - Optional accelerated multi-scale deformable attention (`ms_deform_attn`
-  slot): with `pip install libreyolo[hub-kernels]`, RF-DETR,
-  LibreDeformableDETR, and LibreDINO-DETR run the compiled Apache-2.0 CUDA
-  kernel from `kernels-community/deformable-detr` (forward and backward)
-  instead of the portable `grid_sample` path. Installing the extra is the
+  slot): with `pip install libreyolo[hub-kernels]`, every
+  Deformable-DETR-lineage family runs the compiled Apache-2.0 CUDA kernel
+  from `kernels-community/deformable-detr` (forward and backward) instead of
+  the portable `grid_sample` path: RF-DETR, LibreDeformableDETR,
+  LibreDINO-DETR, LW-DETR, Grounding DINO, RT-DETR, RT-DETRv2, D-FINE,
+  RT-DETRv4, DEIM, DEIMv2, EC and OV-DEIM. Installing the extra is the
   opt-in; `LIBREYOLO_HUB_KERNELS=0` disables it. Eager CUDA fp32 only;
   exports always keep the portable path; load or runtime failures fall back
-  with one warning.
+  with one warning. Shapes the slot cannot express (a per-level sampling
+  point count, or `method='discrete'`) also fall back.
+
+- Fused scaled-dot-product attention across the transformer families, using
+  stock torch (no optional dependency). SegFormer, Depth Anything (and
+  MoGe-2), BERT, Grounding DINO, SwinIR and PP-OCR use it by default; ONNX
+  export still traces the primitive-op equation, so exported graphs are
+  unchanged. Families pinned to a byte-exact parity bar (Swin,
+  LibreDINO-DETR's Swin backbone, BiRefNet, FeyNoBG, OWLv2, LW-DETR,
+  SigLIP 2, ZipDepth, MobileSAM) keep manual attention by default and opt in
+  with `libreyolo.kernels.attention.set_fused_attention(model)`, which trades
+  byte-exact agreement with upstream for the fused kernels. Measured on an
+  RTX 5070 Ti under fp16 autocast: 1.77x on Swin window attention, 3.74x on
+  OWLv2 vision attention. See `docs/kernels.md`.
 
 - `val_loss=True` extended from the `g0` flagships to **every trainable
   family** (`g0`, `g1` and `g2`), across four tasks rather than detection
--- a/docs/kernels.md
+++ b/docs/kernels.md
@@ -14,9 +14,11 @@ dependency is never an error, only a fallback.
 - `kernels/quant/execute/`: finalized-only real-precision paths. No backward,
   real hardware: the `fp8_gemm` tensor-core GEMM (`torch._scaled_mm`), its
   fused Triton prologue/epilogue, and the packed-weight unpack kernels.
-- `kernels/attention/`: attention ops shared across model families. Currently
-  the `ms_deform_attn` slot (multi-scale deformable attention) consumed by
-  the Deformable-DETR-lineage families.
+- `kernels/attention/`: attention ops shared across model families. The
+  `ms_deform_attn` slot (multi-scale deformable attention) consumed by the
+  Deformable-DETR-lineage families, plus `sdpa.py`, which is policy rather
+  than a kernel: it records when a family may hand its attention to torch's
+  fused `scaled_dot_product_attention` and provides the opt-in switch.
 - The reference implementations stay in `libreyolo/quant/fake_quant.py` and
   `libreyolo/quant/packing.py`: `quant/` defines what the numbers mean,
   `kernels/` makes them fast. `packing.py` never has variants because it is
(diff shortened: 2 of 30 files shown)


EHxuban11:
Both findings reproduced and fixed. Thanks — finding 1 in particular is a real gap I created and my PR description overclaimed on it.

### Finding 1: SDPA in jit.trace-captured graphs — fixed (01335728)

Reproduced: tracing SwinIR's `WindowAttention` gave a graph with `aten::scaled_dot_product_attention` and no `aten::softmax`.

Added `manual_attention_required()` (`is_in_onnx_export() or torch.jit.is_tracing()`) and routed all 16 swapped call sites through it. After: the same trace contains `aten::softmax` and no SDPA, and eager inference still runs SDPA.

Took the two exclusions as recommended, and verified the `torch.export` half rather than assuming it:

```
raw torch.export graph            : SDPA present
after run_decompositions()        : SDPA absent      <- Core AI path
```

ExecuTorch goes through `to_edge_transform_and_lower`, which lowers to the Edge dialect via the same core-ATen decomposition. So neither torch.export backend needs the gate, and leaving `torch.compiler.is_compiling()` out keeps `torch.compile` on the fused path.

Also corrected the overclaim: the PR description, `docs/kernels.md` and the CHANGELOG said "exported graphs are unchanged", which was only true for ONNX. They now name the capture kinds explicitly and record why the `ms_deform_attn` slot's rule is deliberately stricter.

New test traces every rewired module and asserts no SDPA in the graph — that is what would have caught this.

### Finding 2: dishonest `set_fused_attention` count — fixed (99e88b79)

Reproduced exactly: `vit` and `deit` both reported `1 module(s) switched` with output unchanged. They are the only two vestigial flags in the repo; `depth_anything3`'s is genuinely read.

Both now honor the flag, default behavior unchanged (max diff after opting out: 7.4e-08 vit, 8.9e-08 deit — same equation, different accumulation order).

**One correction to the review:** the suggested "two-line change in vit/deit forwards; they already contain both code paths" holds for `vit` but not `deit` — `deit`'s `Attention` has *no* manual path and calls SDPA unconditionally, including during ONNX export. That is deliberate: `export/onnx.py:47` lists `deit` in `_requires_onnx_opset17` precisely so its ONNX graph can carry `aten::scaled_dot_product`. So `deit` needed a manual branch written for it, and I gated it on `self.fused_attn` alone with no export condition, so the default keeps today's exact behavior.

For the same reason neither vit nor deit uses the shared `manual_attention_required()`: they are outside the campaign and their jit.trace artifacts were validated *with* SDPA in the graph. Switching them would change validated artifacts for no benefit to this PR. Both gates carry a comment saying so. Happy to align them in a follow-up if you'd rather have one rule everywhere.

### Verification

- Full suite after the fixes: **5033 passed, 0 failed** (was 4989 + 2 flakes; the two known flaky tests happened not to fire this run).
- GPU re-run on the 5070 Ti: all 28 module/mask combinations still agree in fp32 and fp16 autocast. Speedups unchanged — Swin 2.08x, OWLv2 3.77x.
- `ruff` on changed files: back to the 5 pre-existing findings, none new.

### On the minor notes

Both correct, both left alone. The `torch.device("cuda")` vs `"cuda:0"` duplicate-cache-entry point is real — worst case a couple of extra 2-int64 tensors in a 32-slot cache. The double `resolve()` is one cached dict lookup.

EHxuban11:
Conflicts resolved — PR is `MERGEABLE` against `dev` now (bd590b49).

#712 merged, so I retargeted this from `feat/kernels-registry-lift` to `dev` and merged `dev` in. Two conflicts, both from #712's final commit e7ceee9c:

- **`ms_deform_attn.py`** — the real one. That commit added an `is_exporting` guard to `maybe_ms_deform_attn`'s capture check; this PR had refactored the same block into `ms_deform_attn_available()`. Kept both: the helper now carries `is_exporting` alongside the other three, and `maybe_ms_deform_attn` delegates to it. The `_not_exporting` fallback came across unchanged.
- **`CHANGELOG.md`** — both sides added prose to the same `ms_deform_attn` bullet. Merged: the Hub commit-pin note and the fall-back-shapes note now sit in one entry.

`docs/kernels.md` auto-merged cleanly. Nothing in #712's RF-DETR rewiring (moving the slot consult from the core into `MSDeformAttn.forward`) conflicts with this PR — the eight families here consult the slot from their own cores.

**One thing I checked and did *not* act on.** #712 adding `is_exporting` to the MSDA slot made me reconsider whether the SDPA gate should take it too — it would have let `torch.export` capture the manual equation directly instead of relying on decomposition. Measured on torch 2.11 first:

```
mode               is_compiling  is_exporting  manual_required
export-nonstrict   True          True          False
compile            True          True          False
```

`is_exporting()` is True under plain `torch.compile` as well, so it cannot single out `torch.export` — gating on it would have cost the compiled path its fused kernels, exactly what the review warned about. `manual_attention_required()` stays ONNX-or-jit-trace. The docstring and `docs/kernels.md` now record the measurement instead of the weaker "cannot distinguish" claim I had written.

(It is a free addition on the MSDA slot, which refuses under `is_compiling()` anyway and has no compiled-inference story to protect. Correct there, wrong here.)

**Verification after the merge:** full suite **5033 passed, 1 failed** — same pass count as before the merge, and the one failure is the pidnet/ncnn flake documented above (fails ~1 run in 6 alone on `dev`). `tests/unit/kernels/` 184 passed, including #712's new hub-pin tests. `ruff` on the full merged diff: the same 5 pre-existing findings, none new.

PR #747, opened 2026-08-09
Port V-JEPA 2.0 as the vjepa2 video embedding and classification family (both milestones)

EHxuban11:
Ports **V-JEPA 2.0** into LibreYOLO as the `vjepa2` family: a self-supervised video encoder under the generic `embed` task, plus `classify` for its released attentive probes.

This covers **both milestones of the handoff**: the encoder family and embedding contract, the attentive-probe classifier, and the frozen-probe trainer. All eight canonical artifacts are converted, parity-checked and published. Everything claimed here was actually run.

## Code provenance

Two upstream sources, both permissive:

| Surface | Upstream | Pinned at | License |
| --- | --- | --- | --- |
| Encoder / pooler / classifier implementation (`libreyolo/models/vjepa2/nn.py`) | [huggingface/transformers](https://github.com/huggingface/transformers) | tag `v5.1.0`, commit `3fa4da70f5d1da3ab1a8304bd2339eab7004b157`, file `src/transformers/models/vjepa2/modeling_vjepa2.py` | **Apache-2.0** |
| Semantic upstream: architecture behaviour, preprocessing, attentive-probe recipe, independent oracle | [facebookresearch/vjepa2](https://github.com/facebookresearch/vjepa2) | commit `204698b45b3712590f06245fbfba32d3be539812` | **MIT** |

`nn.py` is adapted from the Apache-2.0 Transformers file. It is **not** relicensed as MIT: the Apache-2.0 notice and terms are preserved verbatim in `libreyolo/models/vjepa2/NOTICE`, and the file carries a provenance header. Everything else in this PR is original work written for it.

The three files that `facebookresearch/vjepa2` marks Apache-2.0 rather than MIT (`src/datasets/utils/video/randaugment.py`, `randerase.py`, `worker_init_fn.py`) are **not** adapted, imported or vendored.

No GPL, AGPL, LGPL, non-commercial or unknown-license code is included in any form.

**Weights** are licensed per artifact and deliberately not collapsed into a family-wide claim: the two ViT-g encoders are Apache-2.0, the other six artifacts are MIT. Each rehosted repo ships its own `LICENSE` and `NOTICE`, and `weights/LICENSE_NOTICE.txt` records all eight with exact source revisions.

**Datasets**: none vendored. Something-Something V2 (research-use agreement) and Diving48 (redistribution terms not clearly established by the selected official source) are named for provenance only and are never mirrored, bundled or auto-downloaded.

## Parity: exact, against a pinned oracle

The gate is `max_abs_diff == 0.0` on fixed float32 CPU inputs against **unmodified `transformers==5.1.0`**.

The pin matters and is enforced in code: later Transformers releases refactor the V-JEPA 2 rotary helper and the attention dispatch, so a newer install is not a valid oracle for this gate. `weights/parity_vjepa2.py` refuses to run against any other version.

| Artifact | Frames | Tokens | Token diff | Pooled diff |
| --- | ---: | --- | ---: | ---: |
| `l256` encoder (released weights) | 64 | (1, 8192, 1024) | **0.0** | **0.0** |
| `l256` encoder | 16 | (1, 2048, 1024) | **0.0** | **0.0** |
| `h256` encoder (released weights) | 16 | (1, 2048, 1280) | **0.0** | **0.0** |
| `g256` encoder (released weights) | 8 | (1, 1024, 1408) | **0.0** | **0.0** |
| `g384` encoder (released weights) | 8 | (1, 2304, 1408) | **0.0** | **0.0** |
| `l256` SSv2 probe, single-view logits | 16 | (1, 174) | **0.0** | argmax matches |
| `l256` Diving48 probe, single-view logits | 32 | (1, 48) | **0.0** | — |
| `g384` SSv2 probe, single-view logits | 64 | (1, 174) | **0.0** | — |
| `g384` Diving48 probe, single-view logits | 32 | (1, 48) | **9.54e-07** ⚠️ | — |

**All four encoder sizes pass exactly**, covering all three architecture widths (L=1024, H=1280, g=1408). Strict loads: 388 tensors / 303.9M params (l256), 516 / 631.6M (h256), 644 / 1012.2M (g256 and g384), 453 tensors (L probes), 709 (g384 probes). Only `predictor` is dropped, and it is dropped as a **named asserted set** rather than a substring filter, so a changed upstream layout fails loudly instead of silently discarding tensors.

Round-trip through the public API is also exact: the converted checkpoint loaded via `LibreYOLO(path)` reproduces `0.0` against the oracle.

Frame reversal moves the l256 tokens by 47.39, so the clip path genuinely preserves time rather than averaging it away.

### The one artifact that does not meet the gate

**`LibreVJEPA2g384-cls-diving48` misses `== 0.0`** and I am not going to round
that away. It differs from the oracle by `9.5367431640625e-07`. Everything I
can establish about it:

- That value is **exactly one float32 ULP** at the logit magnitude involved
  (`max|logit| = 8.6712`, so ULP = 2³·2⁻²³ = 9.537e-07). Relative error is
  **1.1e-07, just under float32 epsilon** (1.19e-07) — the two results are
  equal to within float32 representational precision.
- My model is **self-deterministic**: the same input twice gives `0.0`, and a
  contiguous copy of the input gives `0.0`. So it is not run-to-run noise.
- It is **independent of clip length** — the same 9.54e-07 at 4, 8, 16 and 32
  frames. So it is not sequence-length-dependent kernel dispatch, and it does
  not accumulate through the encoder; it is confined to the output.
- **`argmax` is identical**, and stable across runs.
- The sibling `g384` SSv2 probe — *same architecture, same code path, same
  conversion* — is exactly `0.0`, which rules out a structural mapping error
  in the g384 probe graph.

My read is a last-bit rounding difference in the final projection for this
one artifact's weights. I could not reduce it to zero, and I would rather
surface it than quietly widen the gate: `weights/parity_vjepa2.py` still
asserts `== 0.0`, so this artifact fails that assertion by design. A reviewer
who wants it green should decide whether a 1-ULP tolerance is acceptable for
published probe logits, rather than have me make that call silently.

## Three upstream details that turned out to be load-bearing

Each of these still *builds* a model but breaks a released checkpoint or the numerics. All three are commented at the source and locked by tests:

1. The rotary tables are built with `repeat` (the half-width table concatenated with itself) while the rotated companion pairs **adjacent** channels. The conventional interleaved or half-split rotary is not equivalent here.
2. The attentive pooler's MLP ratio is fixed at **4.0** and does not follow the encoder ratio, which is 4.363636... for the g sizes. Using the encoder ratio builds `fc1` at 6144 instead of 5632 and fails the strict load of every released probe.
3. The pooler's cross-attention has **no output projection**, unlike its self-attention.

I got (2) and (3) wrong on the first pass; strict loading caught both.

## Contracts

- **Public vector** is a documented LibreYOLO *pooling contract*: `normalize(tokens.mean(dim=1), dim=-1)`, returned as float32 `(B, D)`. Upstream designates no global retrieval vector, and **no retrieval benchmark is claimed** for it.
- **Native tokens** are exposed only via `model.embed_tokens(...)` as `(B, T', H', W', D)`, never through the 2D `Embeddings` container.
- **Video**: one result per sampled clip, one clip by default. Every other family keeps its per-frame result cardinality — `VIDEO_EMBED_MODE = "frames"` remains the `BaseModel` default and there is a no-regression test.
- **Images** are accepted as an explicitly documented single-frame input: a static appearance representation, *not* a motion one.
- **Artifact matrix** is enforced beyond the filename regex, because the regex parses combinations that no released artifact provides: `-embed` takes no dataset variant, a published `-cls` requires one of the four released size/variant pairs, and a bare `LibreVJEPA2l256.pt` is not canonical.
- **Training**: the frozen attentive-probe recipe runs behind the normal `model.train(data="video_dataset.yaml")` API. Self-supervised `embed` training, predictor pretraining and full encoder fine-tuning (`freeze=0`) each reject *before* a dataset is constructed, with a message pointing at the supported path.
- **Probe geometry** is recovered from checkpoint metadata: the SSv2 L/256 probe is a 16-frame artifact and must not silently run at the encoder's 64 frames.

## Training (milestone 2)

The encoder is frozen and held in eval mode; only the three-layer attentive
pooler and the linear classifier are optimized, with cross-entropy and
best-by-top-1 selection.

The dataset is **entirely user-supplied**. Nothing downloads, mirrors or
bundles SSv2, Diving48, Kinetics or any other corpus, and a test asserts
validation makes no network call. Manifests are validated in full *before the
first epoch* rather than failing on the first bad batch: missing files,
non-integer or out-of-range labels, non-contiguous `names`, and classes named
but never used are all errors. Paths containing spaces parse (the class id is
read from the last field; tab and quoting also work).

Training sampling is a random temporal crop, validation is deterministic, the
collate returns `(B, F, C, H, W)`, and the forward rejects anything not 5D so
time can never fold into batch silently.

Evidence on synthetic motion where **the label is the direction of travel** —
a model that ignored frame order could not beat chance:

```
encoder fully frozen and in eval mode        True
only pooler.* and classifier.* trainable     33 tensors
optimizer parameter set disjoint from encoder True
loss           1.0591 -> 0.0073
top-1 accuracy 1.000  (chance = 0.5)
```

The convergence test is seeded: asserting convergence on a randomly
initialized probe is otherwise a coin flip, and it did fail once before I
pinned the seed.

## Exports

Fixed-frame 5D graphs with dynamic batch. l256 at 8 frames:

| Format | Result |
| --- | --- |
| ONNX | input `images ['batch', 8, 3, 256, 256]` to output `['batch', 1024]`; diff **1.41e-05**; output unit-norm; frame reversal moves the row by 7.9e-03 |
| TorchScript | diff **1.42e-05**, same temporal delta |

Parity is measured by driving the runtime directly with a 5D clip.
`LibreYOLO(exported_path).predict(...)` is **not** supported for this family
yet and now raises an actionable error rather than an opaque rank mismatch —
see the review-findings section.

ONNX auto-opset moves to 17 for this family because the attention lowers to `aten::scaled_dot_product_attention`, which ONNX supports only from opset 14. NCNN and TFLite are recorded **blocked** (both toolchains assume rank-4 image tensors and have no rank-5 path). **TensorRT and OpenVINO are left at `available`, not `validated`, because I did not run them** — no column is ticked that was not measured.

## Weights published

Uploaded to the LibreYOLO org, each with the exact 5-file contract (verified by HTTP 200 and file listing):

- `LibreYOLO/LibreVJEPA2l256-embed` (MIT)
- `LibreYOLO/LibreVJEPA2h256-embed` (MIT)
- `LibreYOLO/LibreVJEPA2g384-embed` (**Apache-2.0** — exercises the mixed-licence path)
- `LibreYOLO/LibreVJEPA2l256-cls-ssv2` (MIT)
- `LibreYOLO/LibreVJEPA2l256-cls-diving48` (MIT)
- `LibreYOLO/LibreVJEPA2g256-embed` (**Apache-2.0**)
- `LibreYOLO/LibreVJEPA2g384-cls-ssv2` (MIT)
- `LibreYOLO/LibreVJEPA2g384-cls-diving48` (MIT)

**All eight canonical artifacts are published**, each verified to contain
exactly the five contract files with the correct per-artifact licence (2
Apache-2.0 encoders, 6 MIT).

Cleared-cache auto-download verified from a directory with no `weights/`, on both filename shapes:

- `LibreVJEPA2l256-embed.pt` — downloaded 1159.3 MB, loaded, produced tokens `(1, 4, 16, 16, 1024)`.
- `LibreVJEPA2l256-cls-ssv2.pt` — downloaded 1348.3 MB and recovered `variant=ssv2`, `nc=174`, `frames=16` and the real SSv2 class names from checkpoint metadata.

## Integration

CHANGELOG `Unreleased / Added`; one README row (a new **Video embeddings** row — V-JEPA 2 embeds a clip while the existing Embeddings entries embed an image, and merging them would hide the distinction the family exists to make); `MODEL_GROUPS` enrollment in `g2`; regenerated `reports/export_inventory.json` (+25 lines, nothing dropped) and `docs/export_support.md`; nomenclature rows plus the canonical filename matrix; per-artifact weight licences; pytest `vjepa2` marker; upload-skill whitelist extended to all eight names.

e2e registration deliberately avoids `MODEL_CATALOG` (its COCO mAP and training gates fail a clip embedder by construction) and `GENERAL_NIGHTLY_INFERENCE_MODELS` (detect-only), so **the general-nightly count in `docs/testing.md` is unchanged**.

## Tests

85 vjepa2 tests pass across `tests/unit/test_vjepa2.py` and `tests/unit/test_vjepa2_training.py` (including a regression test per review finding), and 262 pass across the vjepa2, registry, export-support, inventory, nomenclature and capability-language suites.

CI is fully green on `3d89621`: the complete unit suites on ubuntu, macOS and Windows, plus `provenance`, the distributed job, all packaging jobs and Greptile.

## Not in this PR

- **Independent parity** against the official MIT implementation. The Transformers oracle is exact; the second, independent oracle is still owed.
- Reducing the `g384-cls-diving48` 1-ULP residual to zero (see above).
- Backend 4D-vs-5D input-validation hardening, and the UI smoke.

## Review findings (all three were real, all three fixed)

Greptile ran twice and raised five findings in total. **Every one was a
genuine bug on a public path my parity work never exercised**, and all five
are fixed. The parity scripts drive the module directly, so `predict()` was
never run end to end until the first review pointed at it.

First pass, fixed in `3d89621`:

1. **`_preprocess` broke the inference contract.** The shared runner unpacks a
   4-tuple; this family returned a bare tensor, so `model.predict()` failed
   before the forward pass. The parity scripts drove the module directly and
   `embed_tokens()` has its own path, so nothing caught it.
2. **`VIDEO_EMBED_MODE` had no runtime consumer.** A video still took the
   per-frame path, turning each decoded frame into its own static repeated
   clip. The generic runner now honours clip mode: decode once, one result per
   sampled clip, with unbounded sources refused rather than buffered and a
   frameless video treated as an error rather than a black clip.
3. **Exported graphs vs backend rank.** Confirmed the exact failure
   (`Invalid rank for input: images Got: 4 Expected: 5`). That path is not
   wired for clips, so it now raises an actionable error naming the working
   alternatives, and the export support entry states the limitation instead of
   implying the reload path works.

Second pass, fixed in `8e2b420`:

4. **`embed_tokens` passed the whole 4-tuple to the encoder** — a regression I
   introduced in the fix for (1), and one my earlier check could not catch
   because it ran before that change.
5. **`save=True` / `show=True` silently ignored in clip mode.** A clip
   collapses the video to one result, so there is no annotated frame stream to
   write; the flags are now refused with an explanation instead of dropped.

Verified end to end on a generated 90-frame motion clip:

```
classify probe   -> 1 result, probs (48,)
embed            -> 1 result, (1, 1024), unit norm
embed_tokens     -> (1, 1, 16, 16, 1024)
save=True        -> refused with the explanation
```

## Reviewer notes

- The `greptile` CLI is **not installed** on this machine, so no local pre-push review was run; the bot review on this PR was the first Greptile pass. Its three findings are addressed above.
- A real defect is fixed here that is worth a wider look: `BaseModel.__init__` records the weights path but does not load from it, so this family initially loaded a **randomly initialized** encoder while every parity script that loaded the module directly still passed. Other recently added families may have the same hole.

V-JEPA 2 is added as a native video embedding and attentive-probe classification family, including frozen-probe training and fixed-frame exports. The token extraction repair remains incomplete for supported still-image inputs.
- Adds four encoder configurations and four released attentive-probe artifacts.
- Adds clip-aware inference, preprocessing, dataset validation, and frozen-probe training.
- Adds ONNX and TorchScript export contracts for fixed-frame 5D inputs.
- Extends model registration, documentation, artifact metadata, and focused tests.
</details>

<details><summary><h3>Confidence Score: 4/5</h3></summary>

The PR is not yet safe to merge because still-image calls to embed_tokens reach the five-dimensional encoder with a four-dimensional tensor and fail.

The prior tuple-handling repair now extracts the tensor correctly, but embed_tokens bypasses the ordinary forward path's image-to-static-clip expansion, leaving a supported public input path broken.

**Files Needing Attention:** libreyolo/models/vjepa2/model.py
</details>

| Filename | Overview |
|----------|----------|
| libreyolo/models/vjepa2/model.py | Implements the public V-JEPA 2 wrapper and clip sampling, but embed_tokens still sends supported 4D image inputs directly to the 5D encoder. |
| libreyolo/models/vjepa2/nn.py | Adds the native encoder, attentive pooler, and classifier implementation with an explicitly five-dimensional video input contract. |
| libreyolo/models/base/inference.py | Adds whole-clip dispatch for embedding and classification, including bounded sampling and explicit output-option handling. |
| libreyolo/export/exporter.py | Adds fixed-frame 5D export construction and normalized embedding output wrapping. |
| libreyolo/backends/base.py | Replaces opaque exported-runtime rank failures with an actionable unsupported-path error. |

</details>

```mermaid
flowchart TD
  A[Public V-JEPA 2 source] --> B{API}
  B -->|predict| C[_preprocess]
  C --> D{Tensor rank}
  D -->|4D image| E[_forward expands to 5D static clip]
  D -->|5D clip| F[Encoder]
  E --> F
  B -->|embed_tokens| G[_preprocess and tuple unpack]
  G --> H[Direct encoder call]
  H -->|5D clip| I[Native token grid]
  H -->|4D image| J[Rank error]
```
</details>

<a href="https://app.greptile.com/api/ide/cursor?prompt=%23%23%23%20Issue%201%0Alibreyolo%2Fmodels%2Fvjepa2%2Fmodel.py%3A452%0A**Still-image%20token%20extraction%20fails**%0A%0AWhen%20%60embed_tokens%60%20receives%20a%20supported%20still-image%20source%20or%204D%20image%20tensor%2C%20%60_preprocess%60%20returns%20a%204D%20batch%20and%20the%20method%20forwards%20it%20directly%20to%20the%205D-only%20encoder%2C%20causing%20a%20rank%20error%20instead%20of%20returning%20the%20documented%20token%20grid.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&pr=747&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursorDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursor.svg?v=6">Fix All in Cursor</picture></a> <a href="https://app.greptile.com/api/ide/codex?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22libreyolo%2Flibreyolo%22%20on%20the%20existing%20branch%20%22port-vjepa2-embed%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22port-vjepa2-embed%22.%0A%0A%23%23%23%20Issue%201%0Alibreyolo%2Fmodels%2Fvjepa2%2Fmodel.py%3A452%0A**Still-image%20token%20extraction%20fails**%0A%0AWhen%20%60embed_tokens%60%20receives%20a%20supported%20still-image%20source%20or%204D%20image%20tensor%2C%20%60_preprocess%60%20returns%20a%204D%20batch%20and%20the%20method%20forwards%20it%20directly%20to%20the%205D-only%20encoder%2C%20causing%20a%20rank%20error%20instead%20of%20returning%20the%20documented%20token%20grid.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=libreyolo%2Flibreyolo&pr=747&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodexDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodex.svg?v=6">Fix All in Codex</picture></a>

<sub>Reviews (8): Last reviewed commit: ["Merge dev: converge on the Perception En..."](https://github.com/libreyolo/libreyolo/commit/5f701c5871105d027d33f5631af45bd0d82d640b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51647114)</sub>

> Greptile also left **1 inline comment** on this PR.

**Context used:**

- Knowledge Base — [Model zoo, architectures, and inference backends](https://app.greptile.com/libreyolo/-/custom-context/knowledge-base/libreyolo/libreyolo/-/docs/model-zoo.md)
- Knowledge Base — [Model export](https://app.greptile.com/libreyolo/-/custom-context/knowledge-base/libreyolo/libreyolo/-/docs/model-export.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,37 @@ based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Releases
 before 1.4.0 are documented in the
 [GitHub Releases](https://github.com/LibreYOLO/libreyolo/releases) only.
 
+## [Unreleased]
+
+### Added
+
+- **V-JEPA 2** (`vjepa2`), a self-supervised video encoder, under the generic
+  `embed` task plus `classify` for its released attentive probes. Sizes
+  `l256`, `h256`, `g256` and `g384`; canonical checkpoints
+  `LibreVJEPA2<size>-embed.pt` and `LibreVJEPA2<size>-cls-<ssv2|diving48>.pt`.
+  The encoder is a native pure-PyTorch port, so inference and export do not
+  depend on `transformers` at runtime.
+
+  This is the first family to consume a *clip* rather than a frame. A finite
+  video produces one result per sampled clip, with one clip by default;
+  every other family keeps its per-frame video result cardinality. An image
+  is accepted as an explicitly documented single-frame input, which is a
+  static appearance representation and not a motion one.
+
+  The public embedding is a LibreYOLO pooling contract, defined as the
+  arithmetic mean of the final encoder tokens followed by L2 normalization.
+  Upstream designates no global retrieval vector, and no retrieval benchmark
+  is claimed for it. The native spatiotemporal token grid is available
+  separately through `model.embed_tokens(...)` as `(B, T', H', W', D)`.
+
+  Encoder token and pooled-vector parity against unmodified
+  `transformers==5.1.0` is exact (`max_abs_diff == 0.0`) on float32 CPU for
+  `l256` and `h256`, as is single-view logit parity for the `l256` SSv2 probe.
+  ONNX and TorchScript ship fixed-frame 5D graphs with dynamic batch.
+
+  Self-supervised encoder pretraining is not implemented and rejects with an
+  actionable message; the attentive probe is the supported training path.
+
 ## [1.5.0] - 2026-08-09
 
 The largest release so far: 28 new model families, four new tasks (`edge`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,11 +30,21 @@ before 1.4.0 are documented in the
 
   Encoder token and pooled-vector parity against unmodified
   `transformers==5.1.0` is exact (`max_abs_diff == 0.0`) on float32 CPU for
-  `l256` and `h256`, as is single-view logit parity for the `l256` SSv2 probe.
-  ONNX and TorchScript ship fixed-frame 5D graphs with dynamic batch.
+  all four sizes, as is single-view logit parity for three of the four
+  released probes. The `g384` Diving48 probe differs by one float32 ULP
+  (9.54e-07 absolute, 1.1e-07 relative); it is documented rather than papered
+  over, and the parity script still asserts `== 0.0`. ONNX and TorchScript
+  ship fixed-frame 5D graphs with dynamic batch.
 
-  Self-supervised encoder pretraining is not implemented and rejects with an
-  actionable message; the attentive probe is the supported training path.
+  A frozen attentive-probe trainer ships behind the normal
+  `model.train(data="video_dataset.yaml")` entry point. The encoder is frozen
+  and kept in eval mode; only the three-layer attentive pooler and the linear
+  classifier are optimized, and best checkpoints are selected by top-1
+  accuracy. Datasets are entirely user-supplied: manifests are validated in
+  full before the first epoch, and no video corpus is ever downloaded.
+
+  Self-supervised encoder pretraining, predictor pretraining and full encoder
+  fine-tuning all reject with actionable messages before a dataset is built.
 
 ## [1.5.0] - 2026-08-09
 
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ libreyolo predict --model yolo9-t --source screen            # screen capture
 | **Surface normals** | MoGe-2 |
 | **Edges** | DexiNed, TEED |
 | **Embeddings** | LibreFaceEmbedder, CLIP, SigLIP2, DINOv2 |
+| **Video embeddings** | V-JEPA 2 (clip-level embedding and video classification) |
 | **Body mesh** | SAM 3D Body |
 | **Restoration** | NAFNet, Real-ESRGAN, SwinIR |
 | **Background removal** | BiRefNet, FeyNobg |
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ libreyolo predict --model yolo9-t --source screen            # screen capture
 | **Surface normals** | MoGe-2 |
 | **Edges** | DexiNed, TEED |
 | **Embeddings** | LibreFaceEmbedder, CLIP, SigLIP2, DINOv2 |
-| **Video embeddings** | V-JEPA 2 (clip-level embedding and video classification) |
+| **Video embeddings** | V-JEPA 2 (clip-level embedding, plus video classification with a trainable attentive probe) |
 | **Body mesh** | SAM 3D Body |
 | **Restoration** | NAFNet, Real-ESRGAN, SwinIR |
 | **Background removal** | BiRefNet, FeyNobg |
--- a/docs/export_support.md
+++ b/docs/export_support.md
@@ -98,6 +98,8 @@ in preflight.
 | teed | edge | ✓ | ✓ | ✓ | ✓ | ✓ |  |  |  |  | ✓ |  |  |
 | vgg | classify | ✓ | ✓ | available | ✓ | ✓ |  |  |  | available |  |  |  |
 | vit | classify | ✓ | available | available | available | available |  |  |  | available |  |  |  |
+| vjepa2 | embed | ✓ | ✓ | available | available | available |  |  |  |  |  |  |  |
+| vjepa2 | classify | available | available | available | available | available |  |  |  |  |  |  |  |
 | yolo1 | detect | ✓ | ✓ | ✓ | ✓ | ✓ |  |  |  | ✓ |  |  | ✓ |
 | yolo2 | detect | ✓ | ✓ | ✓ | ✓ | ✓ |  |  |  |  |  |  | ✓ |
 | yolo3 | detect | ✓ | ✓ | ✓ | ✓ | ✓ |  |  |  | ✓ |  |  | ✓ |
@@ -369,6 +371,8 @@ A check mark applies only under any constraint listed here.
 - `vgg` / `classify` / `tensorrt`: FP32, batch 1, fixed 224x224 input
 - `vgg` / `classify` / `openvino`: FP32, batch 1, fixed 224x224 input
 - `vit` / `classify` / `onnx`: FP32, fixed 224x224 input
+- `vjepa2` / `embed` / `onnx`: FP32, dynamic batch, fixed frame count / crop / tubelet geometry per graph; ONNX needs opset >= 14 for scaled_dot_product_attention
+- `vjepa2` / `embed` / `torchscript`: FP32, dynamic batch, fixed frame count / crop / tubelet geometry per graph; ONNX needs opset >= 14 for scaled_dot_product_attention
 - `yolo1` / `detect` / `executorch`: ExecuTorch 1.2, XNNPACK, CPU, FP32, batch 1, fixed input shape
 - `yolo1` / `detect` / `tensorrt`: TensorRT 10.16 FP32 with a fixed canvas; YOLO1 requires 448x448
 - `yolo1` / `detect` / `openvino`: fixed export canvas; YOLO1 requires 448x448
@@ -508,6 +512,14 @@ These converter paths are callable with the recorded validation context.
 - `vit` / `classify` / `tensorrt`: The converter path is available, but the project has not yet recorded TensorRT runtime parity for this family and task.
 - `vit` / `classify` / `openvino`: The converter path is available, but the project has not yet recorded OpenVINO runtime parity for this family and task.
 - `vit` / `classify` / `ncnn`: Conversion is implemented; numeric runtime parity has not been recorded for this combination.
+- `vjepa2` / `embed` / `executorch`: Conversion is implemented; numeric runtime parity has not been recorded for this combination.
+- `vjepa2` / `embed` / `tensorrt`: The converter path is available, but the project has not yet recorded TensorRT runtime parity for this family and task.
+- `vjepa2` / `embed` / `openvino`: The converter path is available, but the project has not yet recorded OpenVINO runtime parity for this family and task.
+- `vjepa2` / `classify` / `onnx`: Conversion is implemented; numeric runtime parity has not been recorded for this combination.
+- `vjepa2` / `classify` / `torchscript`: Conversion is implemented; numeric runtime parity has not been recorded for this combination.
+- `vjepa2` / `classify` / `executorch`: Conversion is implemented; numeric runtime parity has not been recorded for this combination.
+- `vjepa2` / `classify` / `tensorrt`: The converter path is available, but the project has not yet recorded TensorRT runtime parity for this family and task.
+- `vjepa2` / `classify` / `openvino`: The converter path is available, but the project has not yet recorded OpenVINO runtime parity for this family and task.
 - `yolo7` / `detect` / `tensorrt`: TensorRT 10.16 FP32 exports and reloads, but the permissively licensed trained checkpoint changes the public top-k class membership.
 - `yolo9` / `detect` / `rknn`: Exact small variants passed RKNN Toolkit2 2.3.2 compilation, RK3588 PC-simulator raw-output gates, and matched post-NMS detections on a real image. Support is limited to YOLO9-t, YOLO9-E2E-t, YOLO-NAS-s, and PicoDet-s; on-device latency and parity have not been measured. Constraint: RKNN Toolkit2 2.3.2, RK3588 PC simulator, vendor floating build, batch 1, fixed square input
 - `yolo9` / `detect` / `coreml`: Conversion is available, but runtime parity requires a macOS runner.
@@ -1233,6 +1245,20 @@ These converter paths are callable with the recorded validation context.
 - `vit` / `classify` / `tflite`: This family and task have not been validated through the ONNX-to-TFLite path.
 - `vit` / `classify` / `coreml`: This family and task are not covered by the family-aware CoreML wrapper.
 - `vit` / `classify` / `coreai`: This family and task have not been validated for Core AI export.
+- `vjepa2` / `embed` / `paddle`: This family and task have not been validated through the ONNX-to-Paddle conversion path.
+- `vjepa2` / `embed` / `mnn`: MNN v1 has no implemented runtime contract for this family and task.
+- `vjepa2` / `embed` / `rknn`: RKNN v1 is limited to the exact simulator-tested detection variants: YOLO9-t, YOLO9-E2E-t, YOLO-NAS-s, and PicoDet-s on RK3588.
+- `vjepa2` / `embed` / `ncnn`: Both toolchains are built around rank-4 image tensors; a rank-5 clip input has no supported conversion path.
+- `vjepa2` / `embed` / `tflite`: Both toolchains are built around rank-4 image tensors; a rank-5 clip input has no supported conversion path.
+- `vjepa2` / `embed` / `coreml`: This family and task are not covered by the family-aware CoreML wrapper.
+- `vjepa2` / `embed` / `coreai`: This family and task have not been validated for Core AI export.
+- `vjepa2` / `classify` / `paddle`: This family and task have not been validated through the ONNX-to-Paddle conversion path.
+- `vjepa2` / `classify` / `mnn`: MNN v1 has no implemented runtime contract for this family and task.
+- `vjepa2` / `classify` / `rknn`: RKNN v1 is limited to the exact simulator-tested detection variants: YOLO9-t, YOLO9-E2E-t, YOLO-NAS-s, and PicoDet-s on RK3588.
+- `vjepa2` / `classify` / `ncnn`: Both toolchains are built around rank-4 image tensors; a rank-5 clip input has no supported conversion path.
+- `vjepa2` / `classify` / `tflite`: Both toolchains are built around rank-4 image tensors; a rank-5 clip input has no supported conversion path.
+- `vjepa2` / `classify` / `coreml`: This family and task are not covered by the family-aware CoreML wrapper.
+- `vjepa2` / `classify` / `coreai`: This family and task have not been validated for Core AI export.
 - `yolo1` / `detect` / `paddle`: This family and task have not been validated through the ONNX-to-Paddle conversion path.
 - `yolo1` / `detect` / `mnn`: MNN v1 has no implemented runtime contract for this family and task.
 - `yolo1` / `detect` / `rknn`: RKNN v1 is limited to the exact simulator-tested detection variants: YOLO9-t, YOLO9-E2E-t, YOLO-NAS-s, and PicoDet-s on RK3588.
--- a/docs/export_support.md
+++ b/docs/export_support.md
@@ -371,8 +371,8 @@ A check mark applies only under any constraint listed here.
 - `vgg` / `classify` / `tensorrt`: FP32, batch 1, fixed 224x224 input
 - `vgg` / `classify` / `openvino`: FP32, batch 1, fixed 224x224 input
 - `vit` / `classify` / `onnx`: FP32, fixed 224x224 input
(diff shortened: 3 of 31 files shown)


EHxuban11:
All three Greptile P1s were real, and all three are fixed in 3d89621. Each was on a public path my parity work never exercised — the parity scripts drive the module directly, so `predict()` was never run end to end until now.

1. **Preprocessing breaks inference contract** — correct. `_preprocess` returned a bare tensor while the shared runner unpacks `(input_tensor, original_image, original_size, ratio)`, so `model.predict()` failed before the forward pass. Now returns the 4-tuple, with the tensor 5D.

2. **Clip mode lacks runtime handling** — correct. `VIDEO_EMBED_MODE` was declared and asserted in a test, but nothing consumed it, so a video took the per-frame path and each decoded frame became its own static repeated clip. `InferenceRunner._predict_video_clip` now honours it: decode once, one result per sampled clip. Decoding and error handling stay shared; temporal sampling is family-local via `sample_clip_indices()`. Unbounded sources are refused rather than buffered, a container that reports no frame count is decoded bounded instead of trusted, and a video yielding no frames is an error rather than a silent black clip.

3. **Exported backends use wrong rank** — correct, and reproduced exactly: `Invalid rank for input: images Got: 4 Expected: 5`. Wiring clip preprocessing through the exported-backend path is more than this PR should carry, so it now raises an actionable `NotImplementedError` naming the working alternatives, and the export support entry states the limitation rather than implying `LibreYOLO(exported).predict()` works. Export parity itself is measured by driving the runtime directly with a 5D clip.

Verified end to end on a generated 90-frame motion clip: `model.predict(video)` returns exactly one result with a unit-norm `(1, 1024)` embedding.

Added one regression test per finding. 54 vjepa2 tests pass; 570 passed / 12 skipped across the vjepa2, registry, export-support, inventory, nomenclature, capability-language and backend suites.

Also since the first review: `g384` encoder parity is exact (`max_abs_diff == 0.0`, 1012.2M params), so all three architecture widths are covered, and `LibreYOLO/LibreVJEPA2g384-embed` is published — the first Apache-2.0 artifact in the family, which exercises the per-artifact licence path.

EHxuban11:
Second pass: both findings were real, fixed in 8e2b420.

1. **`embed_tokens` passed the whole tuple to the encoder** — correct, and it was a regression I introduced in 3d89621 when `_preprocess` started returning the shared 4-tuple. My earlier `embed_tokens` check ran before that change, so nothing re-covered it. Now unpacked, with a test asserting the token-grid shape so the same slip can't recur silently.

2. **`save`/`show` silently ignored in clip mode** — correct. A clip collapses the video to one result, so there is genuinely no annotated frame stream to write or display. Rather than pretend, the flags are now refused with a message pointing at the returned `Results`.

Verified end to end on a generated motion clip:

```
classify probe   -> 1 result, probs (48,)
embed            -> 1 result, (1, 1024), unit norm
embed_tokens     -> (1, 1, 16, 16, 1024)
save=True        -> refused with the explanation
```

57 vjepa2 unit tests pass. The previous commit was fully green on CI (all three OS suites, provenance, Greptile).

Also newly landed since the last comment: the **Diving48 L/256 probe** converts with exact single-view logit parity (`max_abs_diff == 0.0`, 48 classes, 32 frames, real class names from `id2label`) and is published as `LibreYOLO/LibreVJEPA2l256-cls-diving48`. That makes 5 of 8 artifacts live.

EHxuban11:
Third pass: real again, fixed in c4fad8f.

**Unbounded decode when a container reports no frame count** — correct, and it contradicted the clip contract this PR is built on. The fallback accumulated every decoded frame before selecting the 16 or 64 the clip needs, so memory grew with the length of the video. The clip is bounded, so the decode should be too.

It now counts in a first pass with `grab()` (advances without decoding pixels), then re-opens and retains only the frames the clip actually uses. Peak memory is proportional to the clip, not the video.

Verified against a real video with its frame count forced to zero: one result, unit-norm `(1, 1024)` embedding. A regression test asserts the counting pass uses `grab()` and that the old accumulator does not come back.

Running total: **six findings across three passes, all six real, all six fixed.** Every one was on a public path the parity work never exercised, which is a fair signal about where the risk in this port actually sat — the numerics were exact from early on, the plumbing around them was not. 58 vjepa2 tests now pass, up from 47 before the first review.

EHxuban11:
All eight artifacts are now converted, parity-checked and published. Two things a reviewer should look at:

**1. All four encoder sizes pass exactly.** `g256` came in at `max_abs_diff == 0.0` (1012.2M params, tokens `(1, 1024, 1408)`), so L/H/g widths are all covered at 0.0 for both tokens and the pooled vector.

**2. One probe does not meet the gate, and I am flagging it rather than rounding it away.**

`LibreVJEPA2g384-cls-diving48` differs from the oracle by `9.5367431640625e-07`. Seven of eight artifacts are bit-exact; this one is not. What I established:

- That is **exactly one float32 ULP** at the logit magnitude (`max|logit| = 8.6712` → ULP = 2³·2⁻²³ = 9.537e-07). Relative error **1.1e-07, just under float32 eps** (1.19e-07).
- My model is **self-deterministic** — same input twice gives `0.0`, contiguous copy gives `0.0`. Not run-to-run noise.
- **Independent of clip length**: same 9.54e-07 at 4, 8, 16 and 32 frames. So not sequence-length kernel dispatch, and it does not accumulate through the encoder — it is confined to the output.
- **`argmax` identical** and stable.
- The sibling `g384-cls-ssv2` probe — same architecture, same code path, same conversion — is exactly `0.0`, which rules out a structural mapping error in the g384 probe graph.

Best read is a last-bit rounding difference in the final projection for this artifact's weights. I could not drive it to zero. `weights/parity_vjepa2.py` still asserts `== 0.0`, so this artifact **fails that assertion by design** — I did not widen the gate to make it green. Whether a 1-ULP tolerance is acceptable for published probe logits is a maintainer call, not one I should make silently.

Everything else: 8/8 repos live with the exact 5-file contract and correct per-artifact licences (2 Apache-2.0 encoders, 6 MIT). Cleared-cache auto-download verified on both filename shapes, including `LibreVJEPA2l256-cls-ssv2.pt` recovering `variant=ssv2`, `nc=174`, `frames=16` and real SSv2 class names from metadata. CI is green on `c4fad8f` and Greptile's latest pass raised nothing new.

EHxuban11:
**Milestone 2's trainer has landed, so this PR now covers both milestones of the handoff.**

The frozen attentive-probe trainer runs behind the normal `model.train(data="video_dataset.yaml")` API. Encoder frozen and held in eval mode; only the three-layer pooler and the linear classifier are optimized; best-by-top-1.

Evidence on synthetic motion where **the label is the direction of travel** — a model that ignored frame order could not beat chance:

```
encoder fully frozen and in eval mode          True
only pooler.* and classifier.* trainable       33 tensors
optimizer parameter set disjoint from encoder  True
loss            1.0591 -> 0.0073
top-1 accuracy  1.000  (chance = 0.5)
```

The convergence test is seeded — asserting convergence on a randomly initialized probe is otherwise a coin flip, and it did fail once before I pinned it.

Datasets are entirely user-supplied; a test asserts validation makes no network call, and manifests are validated in full before the first epoch (missing files, bad labels, non-contiguous `names`, unused classes). Self-supervised embed training, predictor pretraining and `freeze=0` all reject before a dataset is built.

**CI caught something I had missed**, which is worth noting: adding a `trainer.py` is not a self-contained change — the repo asserts that every family with one implements the `val_loss` gate and appears in the covered set. The probe is an ordinary cross-entropy classifier, so it now reuses `ClassifyValidationLossMixin` rather than reimplementing the gate. Fixed in `819e408`; all 11 checks green again.

**Final state**

- Both milestones implemented; all 8 canonical artifacts converted, parity-checked and published with the 5-file contract and correct per-artifact licences.
- All four encoder sizes at `max_abs_diff == 0.0`; 3 of 4 probes exact, with the `g384` Diving48 1-ULP residual documented rather than hidden (the parity script still asserts `== 0.0` and fails on it by design).
- 11/11 CI checks pass on all three OSes.
- **Greptile: 6 findings across 3 passes, all 6 real, all 6 fixed**, and the last two passes raised nothing new.

Leaving this for human review — not merging.

PR #789, opened 2026-08-14
Hugging Face Hub integration: load user checkpoints, push fine-tuned ones

### LINKED ISSUE #786: Hugging Face Hub integration: load user checkpoints, push fine-tuned ones
## Summary

Make the Hugging Face Hub a two-way street: let users load LibreYOLO checkpoints from *any* Hub repo, and push their own fine-tuned checkpoints to one. Today the Hub is wired in one direction, for one org.

Optional extra, lazy import, no vendored code. `huggingface_hub` is already an optional dependency used by the openvocab, modus, mobilesam, and hub-kernels paths, so for many installs it is already present.

## Why

Three concrete gaps, all verifiable in the current tree:

1. **Loading is hardwired to the LibreYOLO org.** `BaseModel.get_download_url` (`libreyolo/models/base/model.py:745`) builds `https://huggingface.co/LibreYOLO/{name}/resolve/main/{name}.pt`. There is no way to say `LibreYOLO("someuser/their-model")`. A user who fine-tunes on their own data cannot share the result in the way every other Hub user expects.
2. **Publishing is maintainer-only.** There is no `push_to_hub` anywhere in `libreyolo/`. Weights reach the Hub through 12 bespoke scripts in `weights/` (`upload_alexnet_hf.py`, `upload_birefnet_hf.py`, ..., 3,955 lines total), each hand-written per family. Users have no path at all, and the maintainer path does not scale with the family count.
3. **No model cards for user checkpoints.** The checkpoint metadata standard (`docs/checkpoint_schema.md`) already carries family, task, size, class names, and imgsz. That is most of a model card, and it is currently thrown away at publish time.

### Phase 1: load from any repo

```python
model = LibreYOLO("someuser/my-yolo9-hardhats")   # resolves via the Hub
```

- Treat a `owner/name` string as a Hub repo id, resolve the weight file inside it, and load through the existing factory.
- Works for private repos too, which is the half with no workaround today: a checkpoint that is not under the `LibreYOLO` org cannot be loaded at all right now.
- The existing `LibreYOLO/...` behavior stays the default for bare filenames, so nothing changes for current users.
- `verify_downloaded_file` still applies: a third-party repo is not the LibreYOLO mirror and must not silently inherit its trust.

### Phase 2: push

```python
model.train(data="my-data.yaml", epochs=100)
model.push_to_hub("someuser/my-yolo9-hardhats")
```

- Uploads the checkpoint plus a generated model card built from checkpoint metadata: family, task, size, class names, imgsz, and the license.
- `private=True` supported. Never uploads dataset content.

### Phase 3: retire the bespoke scripts

Re-point the 12 `weights/upload_*_hf.py` scripts at the shared code path and delete the per-family duplication.

## Acceptance criteria

- `LibreYOLO("<user>/<repo>")` loads a checkpoint published by `push_to_hub` and predicts identically to loading the same file from disk.
- Round trip on a small fine-tune: train, push to a private repo, load back, byte-identical weights.
- `import libreyolo` must not import `huggingface_hub`; the entry points raise a clear install hint when it is absent.
- The generated model card states the weight license correctly, including the non-commercial cases the project already tracks.
- Private repos: `HF_TOKEN` (or a cached `huggingface-cli login`) is honoured on load, and a private repo without credentials fails with a message that says authentication is missing, not a bare 404.

## Non-goals

- No hard dependency, no vendored code.
- Not a replacement for the existing auto-download of LibreYOLO's own weights.
- No Hub Spaces / inference-endpoint work in the first iteration.

EHxuban11:
Closes #786 (phases 1 and 2).

## What

- `LibreYOLO("owner/repo")` and `LibreYOLO("hf://owner/repo[@rev][/file.pt]")` load checkpoints from any Hub repo. Download goes through the huggingface_hub cache (auth, resume, private repos for free), then the normal safe-load + metadata-validation path.
- `model.push_to_hub("owner/repo")` uploads a v1.0 checkpoint as `model.pt` plus an auto-generated model card (`library_name: libreyolo`, pipeline_tag, tags, classes, optional license + metrics).
- `HuggingFaceHubLogger` / `loggers="hf:owner/repo"` pushes `weights/best.pt` when training ends. Credentials checked at construction so a missing login fails before training, not after.
- Auth errors teach the three fixes: `hf auth login`, `HF_TOKEN`, `token=`.
- Optional extra `libreyolo[hf]`. `import libreyolo` never imports huggingface_hub.

## Rules

- Local paths always win over bare `owner/repo`. Bare form only claimed when extension-less, valid Hub id, nothing on disk. Shadowed refs get an error hint pointing at `hf://`.
- Repo with several checkpoints errors and lists them; pick with `hf://owner/repo/<file>`.
- Non-v1.0 checkpoints are refused on push with instructions to re-save.

## Tested

- 40 new unit tests (parsing, selection, card, push mocking, logger, resolve_loggers).
- Regressions: test_model_factory, test_training_loggers(_extended), test_serialization, test_download all pass.
- Live smoke on Windows: `hf://LibreYOLO/LibreYOLO9t/LibreYOLO9t.pt` downloads and loads (yolo9 t detect 80).

## Not in this PR

- Phase 3 (porting the 12 weights/ upload scripts onto push_to_hub).
- Docs website page (docs/hf_hub.md added in-repo).

## Code provenance

All code in this PR is original, written for this change. No third-party code was copied. huggingface_hub is used through its public API only.

The PR adds optional Hugging Face Hub checkpoint loading and publishing, including automatic publication of training results.
- Resolves bare and explicit Hub references through the Hugging Face cache before normal checkpoint validation.
- Adds model and checkpoint upload APIs with generated model cards.
- Adds a training logger that uploads the best available checkpoint at training completion.
- Adds the optional `hf` dependency group, documentation, and unit coverage.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| libreyolo/utils/hf_hub.py | Implements Hub-reference parsing, checkpoint resolution, metadata-safe model-card generation, authentication guidance, and checkpoint publication; the prior list-form class-name failure is fixed. |
| libreyolo/models/__init__.py | Routes recognized Hub references through the existing safe checkpoint-loading path while preserving local-path precedence. |
| libreyolo/models/base/model.py | Exposes model publication through the BaseModel API. |
| libreyolo/training/loggers/hf_hub_logger.py | Adds an end-of-training logger that verifies access early and uploads the best available checkpoint. |
| libreyolo/training/loggers/__init__.py | Registers and resolves configured Hugging Face logger instances and prefixed logger strings. |
| tests/unit/test_hf_hub.py | Covers reference parsing, resolution, model cards, publishing, authentication failures, logger behavior, and the corrected list-form class-name path. |
| pyproject.toml | Adds huggingface_hub as an optional dependency through the `hf` and aggregate extras. |

</details>

```mermaid
flowchart LR
  HubRef["owner/repo or hf:// reference"] --> Resolve["Resolve through huggingface_hub cache"]
  Resolve --> Factory["LibreYOLO factory"]
  Factory --> Validate["Safe load and metadata validation"]
  Validate --> Model["LibreYOLO model"]
  Model --> Save["Save schema-v1.0 checkpoint"]
  Save --> Card["Generate model card"]
  Card --> Upload["Upload model.pt and README.md"]
  Training["Training completion"] --> Logger["HuggingFaceHubLogger"]
  Logger --> Best["Select best.pt or last.pt"]
  Best --> Card
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Fix edge cases found by adversarial test..."](https://github.com/libreyolo/libreyolo/commit/19b54cd99339161032dbbb2d415dc295661ce7fb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53519532)</sub>

**Context used:**

- Knowledge Base — [Model zoo, architectures, and inference backends](https://app.greptile.com/libreyolo/-/custom-context/knowledge-base/libreyolo/libreyolo/-/docs/model-zoo.md)
- Knowledge Base — [Training Pipeline](https://app.greptile.com/libreyolo/-/custom-context/knowledge-base/libreyolo/libreyolo/-/docs/training-pipeline.md)

### DIFF
--- a/docs/hf_hub.md
+++ b/docs/hf_hub.md
@@ -1,0 +1,95 @@
+# Hugging Face Hub integration
+
+LibreYOLO can load checkpoints from any Hugging Face Hub repository and push
+your own fine-tuned checkpoints back, in both cases using the standard
+LibreYOLO metadata schema (v1.0, see `docs/checkpoint_schema.md`).
+
+The integration is optional:
+
+```bash
+pip install libreyolo[hf]
+```
+
+`import libreyolo` never imports `huggingface_hub`; it is only loaded when a
+Hub reference or push is actually used.
+
+## Loading models from the Hub
+
+Any repository that contains a LibreYOLO checkpoint can be loaded directly:
+
+```python
+from libreyolo import LibreYOLO
+
+model = LibreYOLO("someuser/their-finetune")          # bare owner/repo
+model = LibreYOLO("hf://someuser/their-finetune")     # explicit form
+model = LibreYOLO("hf://someuser/repo@main/best.pt")  # pin revision and file
+```
+
+Resolution rules:
+
+- A local file or directory with the same name always wins over the bare
+  `owner/repo` form. Use the `hf://` form to bypass that precedence.
+- The bare form is only treated as a Hub reference when the last segment has
+  no file extension, so paths like `weights/model.pt` keep their meaning.
+- Without an explicit filename, the single `*.pt` (or `*.safetensors`) file
+  in the repo is used. Repos with several checkpoints raise an error listing
+  them; select one with `hf://owner/repo/<filename>`.
+
+Downloads go to the shared `huggingface_hub` cache and are loaded through the
+same safe path as local files: `torch.load(weights_only=True)` plus metadata
+validation. Checkpoints without LibreYOLO metadata fall back to the usual
+legacy or auto-conversion flows with a warning.
+
+Private and gated repos work once you are authenticated (see below).
+
+## Pushing models to the Hub
+
+Any loaded model can be published, together with an auto-generated model card
+derived from its checkpoint metadata (family, size, task, classes, metrics):
+
+```python
+model = LibreYOLO("runs/detect/train/weights/best.pt")
+model.push_to_hub("someuser/my-finetune")                  # public
+model.push_to_hub("someuser/my-finetune", private=True)    # private
+model.push_to_hub("someuser/my-finetune", license="mit",
+                  metrics={"mAP50": 0.62})
+```
+
+The repo is created if missing. The checkpoint is uploaded as `model.pt`, so
+repeated pushes update the same file. Anyone can then load it back with
+`LibreYOLO("someuser/my-finetune")`.
+
+To upload automatically when a training run finishes, use the Hub logger:
+
+```python
+model.train(data="data.yaml", epochs=100, loggers="hf:someuser/my-finetune")
+```
+
+or, with options:
+
+```python
+from libreyolo.training import HuggingFaceHubLogger
+
+model.train(
+    data="data.yaml",
+    loggers=HuggingFaceHubLogger("someuser/my-finetune", private=True),
+)
+```
+
+The logger checks for credentials at construction time, so a missing login
+fails before training starts instead of after hours of training. On train
+end it pushes `weights/best.pt` (falling back to `last.pt`) with the final
+metrics rendered into the model card.
+
+## Authentication
+
+Reading public repos needs no login. Private repos, gated repos, and all
+pushes need a Hugging Face token, resolved the standard way:
+
+1. Run `hf auth login` once (stores a token on the machine), or
+2. set the `HF_TOKEN` environment variable, or
+3. pass `token="hf_..."` to `push_to_hub` / `HuggingFaceHubLogger`.
+
+Pushing requires a token with write scope; create one at
+<https://huggingface.co/settings/tokens>. Error messages repeat these steps
+whenever authentication is the likely cause.
--- a/docs/hf_hub.md
+++ b/docs/hf_hub.md
@@ -35,10 +35,19 @@ Resolution rules:
   in the repo is used. Repos with several checkpoints raise an error listing
   them; select one with `hf://owner/repo/<filename>`.
 
+- A revision passed as `@revision` must not contain `/`; use a commit SHA or a
+  slash-free branch name.
+
 Downloads go to the shared `huggingface_hub` cache and are loaded through the
 same safe path as local files: `torch.load(weights_only=True)` plus metadata
-validation. Checkpoints without LibreYOLO metadata fall back to the usual
-legacy or auto-conversion flows with a warning.
+validation.
+
+Hub checkpoints are identified by their LibreYOLO metadata. Recognized
+upstream checkpoints are still auto-converted, but a file with no metadata
+that cannot be converted is rejected rather than guessed at, because raw
+tensor keys from an arbitrary Hub repo can match an unrelated family and fail
+much later with a confusing error. To load such a file anyway, download it and
+pass a local path, which re-enables legacy architecture detection.
 
 Private and gated repos work once you are authenticated (see below).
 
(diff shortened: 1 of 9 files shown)


EHxuban11:
Real end-to-end test on a live account (user Xuban, fine-grained token).

**push_to_hub** — `Xuban/libreyolo-hf-push-test` (private), 8.3MB LibreYOLO9t detect:
- repo created, `model.pt` + `README.md` uploaded, `pipeline_tag: object-detection`, tags `libreyolo/yolo9/detect`, `license:mit`, metrics table rendered.
- round-trip load via bare `Xuban/libreyolo-hf-push-test`: **1338 tensors, 0 bitwise diffs**, names + nc match.
- inference from the Hub-loaded model: 3 person + 1 skateboard on parkour.jpg.
- re-push to same repo updates in place (dedupes, still 3 files, visibility preserved).

**Training auto-upload** — real 1-epoch run, `loggers="hf:Xuban/libreyolo-train-push-test"`:
- pushed `weights/best.pt` on train end; reload gives the fine-tuned head (`nc=1`, `names={0: 'dog'}`) and the run metrics in the card.

**Bug found and fixed** (commit 2e98bade): the logger created the repo **public** by default. An unattended end-of-run upload publishing a model trained on proprietary data is not recoverable, so the logger now defaults to `private=True` while explicit `push_to_hub` stays public-by-default. Regression test added; 41 unit tests pass.

EHxuban11:
Adversarial edge-case pass. 8 bugs found, all fixed in 19b54cd9. 49 unit tests, plus 108 in the surrounding suites.

| # | Bug | Severity |
| --- | --- | --- |
| 1 | `names` as a **list** (valid per schema v1.0) crashed the card builder with `TypeError: list indices must be integers`, so a legitimate checkpoint could not be pushed | high |
| 2 | Foreign Hub repo (a random GPT-2) was misrouted into RF-DETR and failed with "Could not automatically detect LibreRFDETR model size", sending users down a wrong path | high |
| 3 | Card front matter was **injectable** from checkpoint metadata: `model_family: "x\nlicense: proprietary"` forged a `license:` key, i.e. NC weights could be published claiming a permissive license | high |
| 4 | Logger's "fails before training" promise was hollow: it only checked that *a* token existed, so `HuggingFaceHubLogger("google/not-mine")` constructed fine and blew up after the run | high |
| 5 | Offline / no network was reported as "File 'model.pt' does not exist in repo", an actively wrong diagnosis (`LocalEntryNotFoundError` subclasses `EntryNotFoundError`, so it must be caught first) | medium |
| 6 | Missing-file error printed `File 'None' does not exist` whenever the filename had been auto-selected | low |
| 7 | `names` with string keys ordered classes lexicographically in the card (c1, c10, c11, c2) | low |
| 8 | `hf://u/r@refs/pr/1` mis-split into revision `refs` + file `pr/1`; documented as a limitation (revisions may not contain `/`) rather than changing the syntax | low |

Fixes: `_card_names` normalizes list/int-key/string-key names; `_yaml_scalar` quotes anything that is not a plain token; Hub loads now require LibreYOLO metadata (auto-conversion still applies, local paths still get legacy detection); `assert_can_push` resolves the token identity, refuses namespaces the user does not own, and creates the repo as the real proof of write access.

Verified as *not* broken: DDP pushes once (`on_train_end` is behind `is_main_process()`); malicious pickles are refused with no code execution, including through the autoconvert fallback, which is itself hardened with `weights_only=True` + inert stubs that refuse to shadow `os`/`builtins`/`subprocess`; a repo with no loadable weights gives a good message listing the files present.

Live re-check after the fixes: push, tags (`license:mit`, `object-detection`), and round trip still 0 bitwise diffs of 1338 tensors.

